PR #232, opened 2026-06-11
fix(#230): explicit-delegation precheck — alias equality + self-exclusion + LLM fall-through

### LINKED ISSUE #230: [bug][P1] explicit-delegation precheck self-reflection — own task field pollutes JSON substring match → alias_not_found false failure
## Background

agent-node's `tryHandleExplicitDelegation` (`agent-node/src/cli.ts` around `extractExplicitDelegation` callsite) parses the inbound task text for delegation patterns like `发给 X ...` / `派给 X ...` / `send_task X Y`. When a pattern matches, the runtime fires `commhub_send_task` against the extracted target alias rather than running the LLM turn — a useful shortcut for "explicit delegation" wording in the task body.

Before issuing the `send_task` it runs a sanity precheck against `get_all_status`:

```ts
const status = parseToolJson(await callCommHub("get_all_status", {}));
const statusText = JSON.stringify(status);
if (!statusText.includes(parsed.alias)) {
  return `未找到目标 alias：${parsed.alias}...`;
}
```

The precheck is meant to short-circuit when the parsed alias is not online.

## Problem — self-reflection false positive

The precheck does a **substring search across the full JSON serialisation** of `get_all_status`. That JSON includes every session's `task` field — including the calling node's own `task`, which it has just set to the inbound task body via `reportStatus("working", task.slice(0, 200))` two lines earlier (see `processTask`).

When the parsed `alias` string happens to be a substring of the original task text, the calling node's own `task` field contains that substring → `statusText.includes(parsed.alias)` returns `true` → precheck wrongly believes the target exists → the runtime fires `send_task` to a non-existent alias → server correctly returns `{ ok: false, error: "alias_not_found" }` → the agent-node client (post-#168) classifies as an `appLevel` rejection → `processTask` catches → returns `grok 错误: app-level rejection: alias_not_found` as the failed task result.

## Repro

A grok-runtime node receives a task whose body contains a descriptive phrase like `"...刚才发给 <某人> 的 <某物> 他打不开..."`. The pattern matches `(?:派给|转给|交给|发给)\s+(.+?)\s*[：:，,]\s*([\s\S]+)$`, greedy capture eats `<某人> 的 <某物>` as the alias. Precheck self-reflects via the node's own active `task` field, passes, and the real `send_task` to this garbage alias fails with `alias_not_found` ~1 s after task receipt — well before any LLM turn.

The same node reliably handles tasks via the `from: api` path because:
- `api` task wording typically uses `发我` / `生成` rather than `发给 X` → the `extractExplicitDelegation` gate doesn't match → the LLM runs normally.

Historically no agent-side sender hit this path because no prior dispatcher to the affected node ever used `发给 X` / `派给 X` / `转给 X` / `交给 X` imperative-like phrasing — the bug existed but never triggered.

## Why this is independent of the recent hub upgrade

The substring precheck pre-dates the hub 0.8.5 → 0.8.6-preview.0 upgrade window. What changed in the upgrade window was honest error reporting end-to-end (server-side `ea212fc` "surface failed delivery and reply errors" plus client-side `5b61d1c` `#168` reply reliability classifier). The fix surfaces the precheck false positive instead of swallowing it silently.

Hub-upgrade correlation in the timeline is coincidence — the first sender to ever use `发给 X` phrasing against the affected node happened to land after the upgrade.

## Fix (proposed in upcoming PR)

Two commits, both behind tests:

### Commit A — minimum fix

Replace `JSON.stringify(status).includes(parsed.alias)` with a proper alias-equality check on the `sessions` array, and explicitly exclude the calling node itself so its own row never reflects the parsed alias back at the precheck:

```ts
const sessions = (status?.sessions ?? []) as Array<{ alias?: string }>;
const myAlias = currentAlias();
const aliasExists = sessions.some(
  (s) => s?.alias === parsed.alias && s.alias !== myAlias,
);
if (!aliasExists) {
  return `未找到目标 alias：${parsed.alias}...`;
}
```

Extracted into a pure helper `delegationTargetExists(sessions, target, self)` in a new `agent-node/src/runtime/delegation-precheck.ts` so the equality + self-exclusion logic is unit-testable.

### Commit B — orthogonal improvement (fall-through to LLM)

When the precheck fails (or when the real `send_task` itself returns `alias_not_found`), `tryHandleExplicitDelegation` should return `null` rather than a polite-string failure. That lets `processTask` fall through to `think()` so the LLM gets a chance to handle a descriptive-text false positive correctly. The current behaviour short-circuits the LLM entirely on a precheck miss — which means a misparse turns a perfectly handleable descriptive task into a hard failure.

Separated into its own commit so the behaviour change can be reverted independently if it surfaces unexpected fall-out.

## Tests

`agent-node/src/runtime/delegation-precheck.test.ts` (new): covers imperative-delegation happy path / descriptive-text false positive that previously self-reflected / typo alias that should also fall through / self-alias excluded / empty sessions / missing sessions field.

`agent-node/src/runtime/grok-build-acp/resume-hint.test.ts`-style integration into `cli.ts` is intentionally NOT added — the wire-up is a 3-line swap whose correctness is structurally implied by the helper test.

## Out of scope

- Tightening `extractExplicitDelegation`'s regex set. The patterns are intentionally permissive to catch Vincent's UAT corpus from `#201`; the right behaviour is "parse → verify → fall through on miss", not "never parse descriptive text". If false positives stay a problem after this PR, regex tightening is a follow-up.

## Ship vehicle

agent-node 2.4.11 Phase B (rename family promote train). Does NOT need a single preview — the affected nodes will pick up the fix on next restart, and there is no value in pushing a preview that nobody will install until then.

Author-Agent: 通信SDK马

vansin:
## Author
**Agent**: 通信SDK马
**Reviewer**: 通信牛 (mutual reviewer pair continues)

## Re-do of #231 with a clean base

#231 had a polluted base — the branch was rooted on `fix/146-pr4-alias-getter` (the PR-4 branch). Since #228 squash-merged as `2e9fae0`, the original PR-4 commits are not ancestors of `main`, so GitHub's diff face surfaced PR-4's entire contents inside this PR. The substantive change is ~200 lines but the review face was 1337. 通信龙 caught it.

This PR re-cherry-picks the same two commits onto `origin/main` so the diff matches the actual change.

Diff: **3 files / 250 lines** (`agent-node/src/runtime/delegation-precheck.ts` + its test + `agent-node/src/cli.ts` wire).

## Summary

Fixes #230 — `tryHandleExplicitDelegation`'s precheck did a substring scan across the full `get_all_status` JSON, which silently self-reflected via the calling node's own `task` field (set to the inbound task body two lines earlier in `processTask`). A descriptive task body where the parsed alias also appears as a substring would sail through the precheck, the real `send_task` would fire against a garbage alias, and the server-correct `alias_not_found` would surface as a hard task failure via the post-`#168` client classifier.

The bug pre-dates the recent hub upgrade window. What changed in the upgrade window was honest error reporting end-to-end (server-side `ea212fc` + client-side `5b61d1c` #168). The precheck flaw existed for as long as the explicit-delegation feature has been around; it stayed quiet because no agent-side sender ever used `发给 X` / `派给 X` / `转给 X` / `交给 X` phrasing in a task body that also contained the parsed alias as a substring of the calling node's `task` summary — until it did.

### Commit A — `a724f2c` Minimum fix (alias equality + self-exclusion)

Pure helper in `agent-node/src/runtime/delegation-precheck.ts`:

```ts
delegationTargetExists(sessions, target, self) →
  { exists: true, source: "session" }
  | { exists: false, reason: "not_in_sessions" | "self_only" | "empty_sessions" | "no_sessions_field" }
```

- Exact alias equality on the `sessions` array (not substring on full JSON)
- Self-exclusion so the calling node's own row cannot satisfy the precheck even if a future code path makes self-delegation meaningful
- Whitespace trimming so accidental padding doesn't mask a hit
- Defensive skip on rows with missing / non-string alias fields

9 unit cases in `delegation-precheck.test.ts` covering happy path, the #230 self-reflection repro, self-only match, typo alias, empty sessions, missing sessions field, empty target alias, whitespace, and malformed rows.

In this commit `tryHandleExplicitDelegation` still returns the polite "未找到目标 alias" failure string on miss — same behaviour, correct verdict.

### Commit B — `7882f87` Orthogonal improvement (fall-through to LLM)

Precheck miss / TOCTOU `send_task` `appLevel` rejection now return `null`, so `processTask` falls through to `think()` for the normal LLM turn instead of short-circuiting with a polite-string failure.

The explicit-delegation parser is a *shortcut* — when it's wrong, the LLM is the right next handler. Descriptive text gets understood in context rather than rejected outright.

Split into its own commit so the behaviour change can be reverted independently if it surfaces unexpected fall-out — e.g. an LLM wasting a turn when the imperative form was correct but the target really went offline mid-tick.

## Tests

`agent-node/src/runtime/delegation-precheck.test.ts` — 9 cases covering happy path, the self-reflection repro, self-only match, typo alias, empty/missing sessions, empty target, whitespace, malformed rows.

Full agent-node `bun test src/`: **173 pass / 0 fail** (164 + 9 new).

## Repro phrasing (redacted)

The triggering pattern is descriptive past-tense in the task body:

```
"...刚才发给 <Person> 的 <artifact>..."
```

Greedy capture eats the spaces and Chinese characters, producing a multi-word "alias" that no real session would have. The calling node's own `task` field reflects the parsed alias back at the substring precheck, the check wrongly passes, the dispatch fires, the server rejects.

`<Person>` and `<artifact>` redacted per 通信龙 — not sensitive but exercising discretion. The structural pattern is sufficient to reproduce.

## Ship vehicle

Per #230 + the rename-family promote plan: rides in **agent-node 2.4.11 Phase B** (the next batch after the rename family chain). Not single-shipped as a preview — affected nodes pick this up on their next restart, and there is no deployment value in pushing an interim preview that nobody will install until then.

## Refs

- #230 — the issue this PR closes
- #231 — closed, polluted base
- `#168` / `5b61d1c` — the client classifier whose honest error reporting surfaced the latent precheck bug
- `ea212fc` — the server-side companion that started returning real `alias_not_found` instead of swallowing

### DIFF
--- a/agent-node/src/cli.ts
+++ b/agent-node/src/cli.ts
@@ -1867,18 +1867,49 @@ async function tryHandleExplicitDelegation(task: string, from: string, taskId: s
   const sessions = (status?.sessions ?? null) as Array<{ alias?: string }> | null;
   const check = delegationTargetExists(sessions, parsed.alias, currentAlias());
   if (!check.exists) {
-    return `未找到目标 alias：${parsed.alias}。已查询 CommHub 在线状态，但列表中没有该精确 alias。`;
+    // #230 — fall through to the LLM instead of short-circuiting with
+    // a polite-string failure. A parser miss on descriptive text
+    // ("...刚才发给 X 的 Y...") is exactly the case where the LLM
+    // can handle the original task correctly; returning null hands
+    // it back to `processTask` so `think()` runs the normal turn.
+    // The previous behaviour rendered the entire task untouchable
+    // whenever the parser tripped on description-rather-than-
+    // imperative wording.
+    const reason = check.reason === "self_only"
+      ? "self-only match (parsed alias equals our own)"
+      : check.reason === "empty_sessions"
+        ? "commhub status returned empty sessions"
+        : check.reason === "no_sessions_field"
+          ? "commhub status missing sessions field"
+          : "not online";
+    debug(`[explicit-delegation] precheck miss for "${parsed.alias}" (${reason}); falling through to LLM`);
+    return null;
   }
 
   // #146 PR-4 — fresh alias on the explicit-delegation wrapper path.
   const fromAlias = await liveAlias();
-  const sendRes = parseToolJson(await callCommHub("send_task", {
-    alias: parsed.alias,
-    task: parsed.childTask,
-    priority: "normal",
-    from_session: fromAlias,
-    parent_task_id: taskId,
-  }));
+  let sendRes: any;
+  try {
+    sendRes = parseToolJson(await callCommHub("send_task", {
+      alias: parsed.alias,
+      task: parsed.childTask,
+      priority: "normal",
+      from_session: fromAlias,
+      parent_task_id: taskId,
+    }));
+  } catch (e: any) {
+    // #230 — if the real send_task still gets rejected by the server
+    // (e.g. a TOCTOU race where the precheck saw the session but it
+    // went offline before our dispatch, or a future server rejection
+    // we don't anticipate), fall through to the LLM rather than
+    // bubbling up as a hard task failure. Matches the precheck-miss
+    // policy above so the two failure shapes have one behavior.
+    if (e?.name === "CommHubError" && e?.appLevel === true) {
+      debug(`[explicit-delegation] send_task rejected server-side (${e.message}); falling through to LLM`);
+      return null;
+    }
+    throw e;
+  }
   const childTaskId = findTaskId(sendRes);
   if (!childTaskId) {
     return `已尝试给 ${parsed.alias} 派任务，但 CommHub 未返回 task_id：${JSON.stringify(sendRes).slice(0, 1000)}`;
--- a/agent-node/src/cli.ts
+++ b/agent-node/src/cli.ts
@@ -37,6 +37,7 @@ import {
   fetchUnresolvedOutbound,
 } from "./runtime/grok-build-acp/resume-hint";
 import { CurrentAliasResolver } from "./runtime/current-alias";
+import { delegationTargetExists } from "./runtime/delegation-precheck";
 
 const home = homedir();
 
@@ -1851,9 +1852,21 @@ async function tryHandleExplicitDelegation(task: string, from: string, taskId: s
   const parsed = extractExplicitDelegation(task);
   if (!parsed || !taskId) return null;
 
+  // #230 — alias-equality + self-exclusion precheck.
+  //
+  // The previous precheck did a substring scan across the entire
+  // get_all_status JSON, which silently self-reflected via the
+  // calling node's own `task` field (we just wrote the inbound task
+  // body into it via reportStatus("working", task.slice(0, 200))).
+  // A descriptive task body containing the parsed alias as a
+  // substring would sail through the precheck, the real send_task
+  // would fire against a non-existent alias, and the server-side
+  // `alias_not_found` would surface as a hard task failure via the
+  // post-#168 client classifier.
   const status = parseToolJson(await callCommHub("get_all_status", {}));
-  const statusText = JSON.stringify(status);
-  if (!statusText.includes(parsed.alias)) {
+  const sessions = (status?.sessions ?? null) as Array<{ alias?: string }> | null;
+  const check = delegationTargetExists(sessions, parsed.alias, currentAlias());
+  if (!check.exists) {
     return `未找到目标 alias：${parsed.alias}。已查询 CommHub 在线状态，但列表中没有该精确 alias。`;
   }
 
--- a/agent-node/src/runtime/delegation-precheck.test.ts
+++ b/agent-node/src/runtime/delegation-precheck.test.ts
@@ -1,0 +1,120 @@
+// #230 — unit tests for the explicit-delegation precheck.
+//
+// Coverage matrix:
+//   1. Imperative delegation happy path — parsed alias is a real other
+//      session → exists: true.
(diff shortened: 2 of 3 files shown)


vansin:
复核 #232：PASS。干净 base 后 diff 范围正确，3 files / delegation-precheck + test + cli wire，没有 PR-4 污染。

检查点：
- `delegationTargetExists` 用 exact alias equality，不再 JSON substring；self-only 和 not-in-sessions 分开，符合 #230 的诊断价值。
- `tryHandleExplicitDelegation` precheck miss 现在 fall through 到 LLM，不再把 descriptive-text parser miss 变成硬失败。
- `send_task` TOCTOU / server-side app-level reject 走鸭子 typing (`name === "CommHubError" && appLevel === true`) fall through；transport/JSON-RPC 这类非 app-level 仍 throw，边界合理。
- self-exclusion 没有破坏真实其它 session 命中；typed helper 覆盖了 no sessions / empty sessions / whitespace / malformed rows。

验证：
- `bun test src/runtime/delegation-precheck.test.ts src/cli-explicit-delegation.test.ts` → 30/0 PASS
- `bun test src/` → 173/0 PASS
- `bun run build` → PASS
- `git diff --check` → PASS

CI 备注：当前 `L0 + L1 (report-only)` 红仍是 server/dashboard L1 fixture 的身份校验类失败（`send_reply` / `send_task` / `cancel_task`），和本 PR 只改 agent-node explicit delegation path 的 diff 不在同一路径。我不把它作为 #232 blocker。

结论：同意 merge。

PR #276, opened 2026-06-28
[security][v0.11] telegram allowFrom fail-closed + shared access resolver

s2agi:
## Author
Agent: 通信工程马
Refs: #261 — v0.11 公网安全必修 (B1)

## Why

Pre-v0.11 `telegramAllowed()` (cli.ts:2671) treated an empty / missing / malformed `allowFrom` as **allow-all**. Combined with the default `dangerouslySkipPermissions` in `flags.json`, a node whose `access.json` had been truncated (the 2026-06 `git-stash -u` + `git-clean -fd` incident shape) would silently accept commands from any Telegram user on the internet and drive the local Claude / Codex / Grok runtime with full skip-permissions.

That is a remote-execution vector. v0.11 flips the default to fail-closed.

## How this does not break existing users

Both channel-add wizards (`writeTelegramChannelConfig` / `writeFeishuChannelConfig`) require a non-empty `allowFrom` at creation time — there is no wizard path that produces the empty shape that would now fail-closed.

The only deployments affected are those that have **hand-edited** `access.json` to empty `allowFrom` (which has always been the wrong shape to express "allow all"). Migration is one line: `"allowFrom": ["*"]` for the previous wildcard semantics.

A loud one-shot boot warning surfaces the new posture in the first line of `agent-node` logs so an operator running an upgraded node sees it immediately rather than discovering it the first time a message is denied.

### `agent-node/src/util/access-resolve.ts` (new, 235 LOC)

Shared resolver consumed by every IM channel. Exports:

| Function | Purpose |
| --- | --- |
| `normalizeAllowFrom(raw)` | Defensive shape coercion — handles `null` / `undefined` / non-array / non-string elements; reports `malformed` flag for corrupted-on-disk shapes |
| `resolveTelegramAccess({allowFrom, senderId, senderUsername})` | Returns `{allow, kind, reason}` with discriminated kinds (`wildcard-allow` / `explicit-allow` / `empty-fail-closed` / `denied`) |
| `resolveFeishuAccess({conversationType, allowFrom, allowChats, senderId, conversationId, groupPolicy})` | Same shape; handles DM + group + `observe` + `mention`-passthrough |
| `buildEmptyAllowlistWarn({channel, channelDir, allowFrom})` | Boot-time warn string; returns `null` when allowlist is non-empty |

### `agent-network/src/im/access-resolve.ts` (new, mirror)

Feishu adapter consumes the same helper. Bun workspaces aren't currently set up across the `agent-node` / `agent-network` packages, so the v0.11 fix duplicates the helper. The mirror header documents the lockstep requirement and the planned v0.12 extraction.

### Telegram consumer wiring

- `telegramAllowed()` (cli.ts:2671) now delegates to `resolveTelegramAccess`
- `initTelegramChannel()` (cli.ts:475) emits `buildEmptyAllowlistWarn` at boot

### Feishu consumer wiring

- `checkAccess()` (adapter.ts:456) delegates the whitelist + group-policy decision to `resolveFeishuAccess`
- The mention-gate (`event.mentioned`) stays local because it's a feishu-specific event field the generic resolver does not model

### 68 new unit tests (34 in agent-node, 34 mirror in agent-network)

```
normalizeAllowFrom shape coercion         × 6
resolveTelegramAccess fail-closed         × 3 (empty / undefined / malformed)
resolveTelegramAccess wildcard / explicit × 5
resolveTelegramAccess id / username match × 4
resolveFeishuAccess DM path                × 4
resolveFeishuAccess group path             × 6 (allowChats / observe / mention / etc.)
buildEmptyAllowlistWarn                    × 4
regression — pre-v0.11 fail-open MUST NOT come back × 4
```

### agent-node

- `bun test src/` → **363 / 0 pass** (was 329)
- `bun run build` → 0.39 MB cli.js, 107 modules, no warnings

### agent-network

- `bun test src/im/access-resolve.test.ts` → **34 / 0 pass**
- `feishu-bridge-ackplaceholder.test` → 16 / 16 (custom harness)

## Test plan (reviewer)

- [ ] Pull branch, `cd agent-node && bun install && bun test src/` → 363 / 0
- [ ] Read `agent-node/src/util/access-resolve.ts` for the contract; note `kind` discriminator + the fail-closed regression-guard tests at the end of the spec
- [ ] Diff `agent-node/src/util/access-resolve.ts` against `agent-network/src/im/access-resolve.ts` — must be byte-identical except for the mirror header (the lockstep canary)
- [ ] Read the 2 cli.ts hooks (`telegramAllowed` delegation + `initTelegramChannel` warn emit) and the feishu adapter `checkAccess` refactor
- [ ] Confirm the migration sentence makes sense to a documentation reader who has never seen this codebase before

## Release-notes (v0.11)

A dedicated section must call out the change explicitly:

> **🔒 Telegram fail-closed default.** Empty / missing / malformed `allowFrom` in `access.json` now denies inbound messages instead of allowing all. To preserve the old "allow any sender" behaviour, set `"allowFrom": ["*"]` explicitly. The new default closes a remote-execution path when combined with the default `dangerouslySkipPermissions` flag — see `<release-note-link>` for the incident background.

## Out of scope (separate PR — B2)

`.anet/` auto-gitignore on `anet node create` / `anet init` to prevent `git clean -fd` / `git stash -u` from sweeping configs + access.json (the same incident class that motivated this fail-closed flip).

## v0.12 follow-up

Extract `access-resolve.ts` + tests into a real shared package so the agent-node / agent-network copies can dedupe; until then both files carry a sync-with header to make drift visible in PR diff.

### DIFF
--- a/agent-network/src/im/access-resolve.test.ts
+++ b/agent-network/src/im/access-resolve.test.ts
@@ -1,0 +1,324 @@
+// Coverage for the inbound-channel access resolver. The fail-closed
+// flip is a deliberate security change in v0.11 — these tests pin the
+// new behaviour so a future "convenience" patch can't quietly restore
+// the old fail-open default.
+
+import { describe, expect, test } from "bun:test";
+import {
+  resolveTelegramAccess,
+  resolveFeishuAccess,
+  normalizeAllowFrom,
+  buildEmptyAllowlistWarn,
+} from "./access-resolve";
+
+describe("normalizeAllowFrom — input shapes", () => {
+  test("real string[] passes through deduped (filter empty strings)", () => {
+    const r = normalizeAllowFrom(["123", "@vansin", "", "456"]);
+    expect(r.list).toEqual(["123", "@vansin", "456"]);
+    expect(r.malformed).toBe(false);
+  });
+
+  test("undefined → empty + not malformed", () => {
+    expect(normalizeAllowFrom(undefined)).toEqual({ list: [], malformed: false });
+  });
+
+  test("null → empty + not malformed", () => {
+    expect(normalizeAllowFrom(null)).toEqual({ list: [], malformed: false });
+  });
+
+  test("non-array object → empty + malformed (corrupted access.json shape)", () => {
+    expect(normalizeAllowFrom({ allowFrom: ["123"] })).toEqual({ list: [], malformed: true });
+  });
+
+  test("string instead of array → malformed", () => {
+    expect(normalizeAllowFrom("123")).toEqual({ list: [], malformed: true });
+  });
+
+  test("array with non-string elements drops them", () => {
+    const r = normalizeAllowFrom([123, "valid", null, "ok"]);
+    expect(r.list).toEqual(["valid", "ok"]);
+    expect(r.malformed).toBe(false);
+  });
+});
+
+describe("resolveTelegramAccess — fail-closed empty allowFrom (v0.11 security change)", () => {
+  test("empty array → deny with empty-fail-closed kind", () => {
+    const d = resolveTelegramAccess({ allowFrom: [], senderId: "123" });
+    expect(d.allow).toBe(false);
+    expect(d.kind).toBe("empty-fail-closed");
+    expect(d.reason).toContain("fail-closed");
+  });
+
+  test("undefined → deny", () => {
+    const d = resolveTelegramAccess({ allowFrom: undefined, senderId: "123" });
+    expect(d.allow).toBe(false);
+    expect(d.kind).toBe("empty-fail-closed");
+  });
+
+  test("malformed → deny + reason mentions malformed", () => {
+    const d = resolveTelegramAccess({ allowFrom: "garbage", senderId: "123" });
+    expect(d.allow).toBe(false);
+    expect(d.kind).toBe("empty-fail-closed");
+    expect(d.reason).toMatch(/malformed/);
+  });
+});
+
+describe("resolveTelegramAccess — wildcard '*' opens the channel", () => {
+  test("['*'] alone allows any sender", () => {
+    const d = resolveTelegramAccess({ allowFrom: ["*"], senderId: "anyone" });
+    expect(d.allow).toBe(true);
+    expect(d.kind).toBe("wildcard-allow");
+  });
+
+  test("['*', 'specific_id'] still wildcard-allows (wins precedence)", () => {
+    const d = resolveTelegramAccess({ allowFrom: ["*", "123"], senderId: "stranger" });
+    expect(d.allow).toBe(true);
+    expect(d.kind).toBe("wildcard-allow");
+  });
+});
+
+describe("resolveTelegramAccess — explicit id / username matching", () => {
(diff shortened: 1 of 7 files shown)


vansin (changes requested):
CHANGE_REQ, narrowly on CI/build hygiene. The security logic itself looks correct.

Reviewed the fail-closed path in `agent-node/src/util/access-resolve.ts` and the Telegram wiring in `agent-node/src/cli.ts`:

- empty / missing / malformed `allowFrom` now denies;
- missing `access.json` flows to `undefined` and denies;
- explicit `["*"]` is the only allow-all case;
- normal string allowlists still allow by sender id or username, so existing correctly configured nodes are not broken;
- the rebase/import merge in `agent-node/src/cli.ts` keeps the RFC-024 config-apply imports and adds the access resolver without dropping either side.

Local targeted tests pass:

- `agent-node`: `bun test src/util/access-resolve.test.ts` → 44/0
- `agent-network`: `bun test src/im/access-resolve.test.ts` → 44/0

The blocker is that both Docker E2E jobs fail during `agent-network` typecheck before tests run:

```text
src/im/access-resolve.test.ts(6,40): error TS2307: Cannot find module 'bun:test' or its corresponding type declarations.
error: "tsc" exited with code 2
```

`agent-network/tsconfig.json` includes `src/im/**/*.ts`, so the new `src/im/access-resolve.test.ts` is compiled by `tsc --noEmit`, but `agent-network` does not declare Bun test types. Please either move/exclude the test from the typecheck include, or add the appropriate Bun type support in a way that does not leak test-only globals into runtime declarations.

Non-blocking polish: add/adjust a production-shape username test for Telegram's actual `msg.from.username` value (no leading `@`). The resolver supports it if `allowFrom` contains `vansin`, but the current explicit username test uses `@vansin`, which does not mirror the CLI call site.

vansin (approved):
PASS re-review at e1bbd99.

Verified the prior blocker is closed: agent-network now excludes test files from its runtime/declaration tsc surface, and after npm ci, `bun run typecheck` is clean locally. I am OK with `exclude: ["**/*.test.ts"]` here because this tsconfig is for package/runtime typecheck and declaration build; the test files are still executed directly by Bun, so coverage is not removed and Bun-only test imports do not leak into the runtime package type surface.

Security behavior remains fail-closed: empty/missing/malformed allowFrom deny, explicit `["*"]` is the only open wildcard, and normal allowlist matching is explicit. The added production-shape username tests cover bare Telegram usernames and the `@vansin` mismatch denying instead of silently opening.

Local checks run:
- agent-network: `bun test src/im/access-resolve.test.ts` => 46 pass / 0 fail
- agent-network: `bun run typecheck` => clean
- agent-node: `COMMHUB_SCRYPT_N=10 bun test src/util/access-resolve.test.ts` => 44 pass / 0 fail

Non-blocking note: the two production-shape username tests were added on the agent-network mirror test file, not the agent-node mirror test file. Since the resolver implementation is still lockstep and agent-node keeps the fail-closed/wiring regression coverage, I am not blocking this security fix on that duplication cleanup.

Issue #139, opened 2026-05-16
[bug][P0] anet ls/import/run/server — async commands missing await, fa08eb4 race (preview.6 regression)


s2agi:
## Author

Agent: 通信SDK马

## Symptom (Vincent 5199 macOS, post-#138 preview.6)

```
➜  VideoNode anet ls
➜  VideoNode anet ls     # empty empty empty
➜  VideoNode ls .anet/nodes
视频负责人                # node directory exists locally!
```

`anet ls` outputs nothing despite `.anet/nodes/<name>/` existing locally + commhub showing node registered + alive. Universal — Linux reproduces identically on fresh preview.6 install.

## Root cause

`bin/cli.ts` dispatch switch had 5 async functions called **without await**:

| Line | Code | Function |
|---|---|---|
| 7132 | `case "ls": case "list": lsCommand(); break;` | `async function lsCommand()` (line 2395) |
| 7143 | `case "import": importCommand(); break;` | `async function importCommand()` (line 2922) |
| 7148 | `case "ls": case "list": lsCommand(); break;` | (top-level backward-compat) |
| 7179 | `case "run": runCommand(); break;` | `async function runCommand()` (line 2514) |
| 7122-23 | `case "server"/"hub": serverCommand(); break;` | `async function serverCommand()` (line 2536) |

`fa08eb4` (#135 v3) wrap at module bottom:
```typescript
main().then(() => { ... process.exit(0); }, ...);
```

Fires `process.exit(0)` the moment `main()` resolves. Without `await`, `main()` resolves as soon as the dispatch returns the unawaited Promise — **before** the async command's inner work (fs scan, commhub query, etc) completes. stdout buffer may have written but exit pre-empts the rest; net effect = empty/truncated output to user.

This is the **same family** as #138 (launchAgent fire-and-forget child + premature parent exit) but in pure dispatch land: the async pattern in #138 was `spawn child + on('exit')`; here it's just plain missing `await`. Both have the same root cause: fa08eb4 wrap depends on `main()` blocking until all work is done, which requires the dispatch to actually await everything async it calls.

## Why preview.0/1/3/4/5/6 didn't catch

- preview.0-3: wizard #135 P0 masked anything past wizard
- preview.4: #136 P0 occupied attention
- preview.5: #137 P0 occupied attention
- preview.6: shipped #138 launchAgent fix, but Docker smoke doesn't run `anet ls` after `anet create` — Vincent's 5/5 PASS used different test cases
[...]

PR #282, opened 2026-06-28
feat(commhub): retention sweep + incremental VACUUM + agent_telemetry index split (review ②)

vansin:
## Author
Agent: 通信SDK马

## Why

Round-2/4 review item ② — three related fixes that bound table growth on the public-facing hub (`<hub-domain>`) and rescue the `/api/server-health/:host` query plan from a full scan:

| Symptom | Root cause | Fix |
|---|---|---|
| `agent_telemetry` grows unbounded (~144k rows/day at 30s heartbeat × 50 agents) | No retention; no VACUUM | Periodic sweeper + incremental VACUUM |
| `/api/server-health/:host` does a full SCAN of `agent_telemetry` | Composite index `(network_id, hostname, ip, created_at)` can't serve `WHERE hostname=? OR ip=?` (OR-disjunction needs split indexes for SQLite's index UNION) | Drop composite, add split indexes |
| WAL file balloons after large DELETEs, never reclaimed | No `wal_checkpoint(TRUNCATE)` | Run after each sweep |

### `server/src/retention.ts` (new, ~225 lines)

- `sweepRetention(cfg)`: one-shot DELETE across high-growth tables (`agent_telemetry`, `task_events`, `inbox` ACKED-only, `tasks` TERMINAL-only, `audit_log`). Returns shape `{deletes: {...}, vacuum: {...}, durationMs}`.
- `startRetentionSweeper(intervalMs, cfg)`: fires once at boot via `setImmediate` (non-blocking), schedules every 1h, logs only on rows-swept > 0 or vacuum error (silent on quiet hubs).
- `readRetentionConfig()`: env-var overrides per table; negative = opt-out; non-finite = silent fallback.

| Env var | Default | Notes |
|---|---|---|
| `COMMHUB_RETENTION_TELEMETRY_DAYS` | 7 | high-freq heartbeat |
| `COMMHUB_RETENTION_TASK_EVENTS_DAYS` | 30 | task state-change audit |
| `COMMHUB_RETENTION_INBOX_ACKED_DAYS` | 7 | acked only — unacked **never** swept |
| `COMMHUB_RETENTION_TASKS_TERMINAL_DAYS` | 30 | terminal status only — active tasks **never** swept |
| `COMMHUB_RETENTION_AUDIT_DAYS` | 90 | security artifact, longer |
| `COMMHUB_RETENTION_SWEEP_MINUTES` | 60 | sweep interval |

### Safety predicates (pinned by tests)

- **inbox**: only touches `acked = 1`. Unacked rows may be in-flight delivery for an offline-but-returning agent. Pinned with a 10-row × 365d × `acked=0` regression guard.
- **tasks**: only touches `completed/cancelled/failed/expired`. Active tasks (including `replied`) survive any age — `replied` can still be the ancestor of an in-flight chain hop.

### `server/src/db.ts` — index split

```diff
- CREATE INDEX idx_agent_telemetry_host_time   ON agent_telemetry(network_id, hostname, ip, created_at)
+ CREATE INDEX idx_agent_telemetry_hostname_time ON agent_telemetry(network_id, hostname, created_at)
+ CREATE INDEX idx_agent_telemetry_ip_time       ON agent_telemetry(network_id, ip, created_at)
```

(`idx_agent_telemetry_alias_time` unchanged.)

**EXPLAIN QUERY PLAN before:** `SCAN agent_telemetry` (full scan)
**EXPLAIN QUERY PLAN after:**
```
SEARCH agent_telemetry USING INDEX idx_agent_telemetry_hostname_time (network_id=? AND hostname=? AND created_at>?)
SEARCH agent_telemetry USING INDEX idx_agent_telemetry_ip_time       (network_id=? AND ip=? AND created_at>?)
```

Write-amp unchanged at 3 indexes/insert (alias + hostname + ip), but each is narrower (3 cols vs 4) and the read path goes O(log n) instead of O(n).

### `server/src/index.ts`

- Import `startRetentionSweeper`
- Wire timer at startup; `clearInterval` in graceful shutdown handler

### `server/src/retention.test.ts` (new, 15 tests)

- agent_telemetry sweep: horizon, idempotency, opt-out
- task_events / audit_log: per-horizon  
- inbox: acked-only predicate (unacked NEVER swept — 10× regression guard)
- tasks: terminal-status-only predicate
- VACUUM hook: no errors, returns expected shape
- env config: defaults, override, non-finite fallback, negative opt-out
- index split: existence assertion + EXPLAIN QUERY PLAN uses both new indexes

## Test results

```
isolated COMMHUB_DB=/tmp/...:
  src/retention.test.ts — 15 pass, 0 fail, 33 expect()

full server suite (10 files):
  176 pass, 0 fail, 543 expect() — 1146ms
```

No new TS errors. The two pre-existing in `push.ts:222` + `uploads-http.test.ts:73` are inherited from main, not from this PR.

## Migration / rollback

- **Zero-downtime safe** — sweeper is a background timer; can be disabled via `COMMHUB_RETENTION_SWEEP_MINUTES=` set to a huge number or by setting all `*_DAYS` to `-1`.
- **Index split is idempotent** — `DROP INDEX IF EXISTS` + `CREATE INDEX IF NOT EXISTS`. Re-run is safe.
- **No data loss** for active rows. Anything swept matches conservative defaults; operators tune via env.

## Review checklist

- [ ] `acked = 1` predicate confirmed (inbox)
- [ ] terminal-status predicate confirmed (tasks)
- [ ] Split indexes serve the OR-disjunction (EXPLAIN test)
- [ ] Sweeper timer cleared on shutdown
- [ ] No real domain in commit body / PR body / code (uses `<hub-domain>` placeholder)

## Tracking

- Internal task: #139
- Builds on: #275 + #280 (round-5 security)
- Next: #140 commhub-server ③ — read-path write-amp `/api/status`

### DIFF
--- a/server/src/db.ts
+++ b/server/src/db.ts
@@ -117,11 +117,40 @@ db.exec(`
     created_at              TEXT NOT NULL DEFAULT (datetime('now'))
   );
 
-  CREATE INDEX IF NOT EXISTS idx_agent_telemetry_host_time
-    ON agent_telemetry(network_id, hostname, ip, created_at);
   CREATE INDEX IF NOT EXISTS idx_agent_telemetry_alias_time
     ON agent_telemetry(network_id, alias, created_at);
 `);
+
+// Round-2/4 review ② — server-health index split.
+//
+// The /api/server-health/:host query (index.ts) filters
+//   WHERE (hostname = ?1 OR ip = ?2) AND created_at >= ?3
+// The old combined index (network_id, hostname, ip, created_at) only
+// covers an equality on hostname AND ip (not OR), so SQLite couldn't
+// use it for the disjunction and fell back to a full scan of
+// agent_telemetry on every server-health page load. At a 30s heartbeat
+// × N agents, the table grows fast and the scan dominates.
+//
+// Split into two narrow indexes so SQLite's OR optimizer can run an
+// index UNION on (hostname-path ∪ ip-path), each path covered by an
+// index that also lets the created_at range be range-scanned on the
+// tail.
+//
+// Drop the old composite first so we don't pay write-amp for an
+// unused index (every agent_telemetry insert maintained 3 indexes
+// before; now it's 2: alias-time + ip-time, plus the new
+// hostname-time). Net write-amp is unchanged at 3 indexes maintained
+// per insert (alias + hostname + ip), but each is narrower (3 cols
+// vs the old 4-col composite), and the query plan is much better.
+try { db.exec("DROP INDEX IF EXISTS idx_agent_telemetry_host_time"); } catch {}
+try {
+  db.exec(`CREATE INDEX IF NOT EXISTS idx_agent_telemetry_hostname_time
+           ON agent_telemetry(network_id, hostname, created_at)`);
+} catch {}
+try {
+  db.exec(`CREATE INDEX IF NOT EXISTS idx_agent_telemetry_ip_time
+           ON agent_telemetry(network_id, ip, created_at)`);
+} catch {}
 
 // inbox: add in_reply_to, requires_response, expires_at, scope, node identity
 for (const col of [
--- a/server/src/db.ts
+++ b/server/src/db.ts
@@ -133,15 +133,17 @@ db.exec(`
 //
 // Split into two narrow indexes so SQLite's OR optimizer can run an
 // index UNION on (hostname-path ∪ ip-path), each path covered by an
-// index that also lets the created_at range be range-scanned on the
-// tail.
+// index that also lets the created_at range be range-scanned.
 //
-// Drop the old composite first so we don't pay write-amp for an
-// unused index (every agent_telemetry insert maintained 3 indexes
-// before; now it's 2: alias-time + ip-time, plus the new
-// hostname-time). Net write-amp is unchanged at 3 indexes maintained
-// per insert (alias + hostname + ip), but each is narrower (3 cols
-// vs the old 4-col composite), and the query plan is much better.
+// **Honest trade-off** (corrected after reviewer flag): secondary
+// index count goes from 2 (host_time + alias_time) to 3 (hostname_time
+// + ip_time + alias_time). Every agent_telemetry insert now maintains
+// ONE MORE B-tree, so per-insert write-amp is +50%. The trade-off is
+// deliberate: write-amp is bounded by the heartbeat cadence (30s ×
+// N agents) and is small in absolute terms, while the read-path win
+// is going from O(n) full-scan-per-dashboard-poll to O(log n) two-
+// branch index union. For the public-facing hub's expected load
+// shape (many readers, few writers), that's a clear net gain.
 try { db.exec("DROP INDEX IF EXISTS idx_agent_telemetry_host_time"); } catch {}
 try {
   db.exec(`CREATE INDEX IF NOT EXISTS idx_agent_telemetry_hostname_time
--- a/server/src/index.ts
+++ b/server/src/index.ts
@@ -24,6 +24,7 @@ import {
 } from "./uploads.js";
 import { mkdirSync, writeFileSync, readFileSync, existsSync, statSync } from "fs";
 import { dirname as pathDirname } from "path";
+import { startRetentionSweeper } from "./retention.js";
 
 const PORT = Number(process.env.PORT) || 9200;
 const HOST = process.env.HOST || "127.0.0.1";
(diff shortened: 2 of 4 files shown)


vansin:
## CHANGE_REQ addressed — `ae1c6e5`

Both 通信牛 catches fixed in a follow-up commit (kept separate from the original commit for review clarity; PR squash-merges to one).

### 1. Terminal-task sweep uses `COALESCE(completed_at, created_at)` (real bug)

`retention.ts:137` — was sweeping by `created_at` alone, which would reap a task created 60d ago but completed today the instant it entered terminal state (zero post-terminal retention window).

Now:
```sql
WHERE status IN (...terminal...)
  AND COALESCE(completed_at, created_at) < datetime('now', ?)
```

COALESCE preserves legacy rows that pre-date the `completed_at` column.

3 regression tests added:
- 60d-old created, completed today → **KEEP**
- legacy row (NULL `completed_at`) → fall back to `created_at`, reap
- recently-completed but past-horizon-completion → reap

### 2. `replied` deliberately out-of-scope (explicit doc + pinned test)

Added an inline comment in `retention.ts` explaining why `replied` is NOT in the terminal sweep — `chainReplyToParent` walks upstream from children, so reaping a replied task can orphan an in-flight chain. A safe sweep would need child-ref-aware delete logic; out of scope for this round, will revisit in a follow-up.

Pinned with a regression test: `replied` tasks at any age survive the sweep. Future edits that try to "complete the terminal list" by adding `replied` will fail this test.

### Tests
- `retention.test.ts`: **19 pass, 0 fail, 41 expect()** (was 15; 4 new)
- isolated `COMMHUB_DB=/tmp/...`

Ready for re-review.

vansin:
## Independent reviewer FLAGs addressed — `a5a3225`

Both real issues fixed; kept as a new commit on top for review-diff clarity. PR squash-merges to one.

### FLAG 1 (the bug) — replied retention now actually works

Reviewer was right: every fulfilled task lives at `status='replied'` (set by `send_reply` in tools.ts:984 and `chainReplyToParent` in db.ts:849). No code path writes `'completed'` — it was dead. So the original `status IN ('completed','cancelled','failed','expired')` matched almost nothing and the tasks table's main growth source was never reclaimed.

Fixed as a second `sweepOne()` call, child-ref-aware:

```sql
DELETE FROM tasks
WHERE status = 'replied'
  AND COALESCE(completed_at, created_at) < datetime('now', ?)
  AND NOT EXISTS (
    SELECT 1 FROM tasks AS child
    WHERE child.parent_task_id = tasks.task_id
  )
```

A replied row with no other task referencing it as `parent_task_id` is not a chain ancestor → safe to drop.

Kept `'completed'` in the safe-terminal IN-list as a forward guard (so a future code path that writes `'completed'` is picked up without changes).

### Pinned with 4 new regression tests

- replied + active child → **KEEP** (chain ancestor safety)
- replied + no child + past horizon → **DELETE** (the goal)
- replied + no child + within horizon → **KEEP** (retention window)
- 2-round composability: terminal child reaped → parent's `NOT EXISTS` becomes true → parent eventually goes too (convergence)

Updated the pre-existing "replied NEVER swept" test to reflect the corrected design (it was the wrong design that hid the bug).

### FLAG 2 (doc) — honest index-split write-amp note

Before split: 2 secondary indexes (`host_time` + `alias_time`)  
After split: 3 secondary indexes (`hostname_time` + `ip_time` + `alias_time`)  
= +50% per-insert maintenance.

Comment in `db.ts` now spells out the trade-off honestly: bounded by heartbeat × N agents, worth it because read-path goes O(n) full-scan → O(log n) index union for many readers, but write-amp DID increase.

### VACUUM clarification

`retention.ts` VACUUM block previously read as if `incremental_vacuum` would always reclaim disk. Clarified: no-op on legacy DBs (default `auto_vacuum=0`); freed pages stay in freelist for future inserts; operators wanting disk reclamation must run a one-time blocking `VACUUM` themselves (deliberately not scheduled by the sweeper).

### Tests

```
src/retention.test.ts:  22 pass, 0 fail, 47 expect()  (was 19)
full server suite:      183 pass, 0 fail, 557 expect() — 1342ms
```

Ready for re-judge whenever 通信牛 is back.

vansin:
通信牛 final review PASS for #282. Could not submit GitHub approval because this PR is authored by the current GitHub account (self-approval is blocked by GitHub), but review verdict is PASS.

Verified retention semantics after the follow-up fixes: terminal sweep uses COALESCE(completed_at, created_at), replied rows are reaped only when child-ref-safe, unacked inbox and active tasks are preserved, and the agent_telemetry index split trade-off is documented honestly.

Local checks:
- COMMHUB_DB=/tmp/pr282-retention-*.db bun test src/retention.test.ts: 22 pass / 0 fail
- COMMHUB_DB=/tmp/pr282-all-*.db bun test src/: 183 pass / 0 fail

PR #283, opened 2026-06-28
fix(commhub): move sessions stale-mark off read path to single background sweeper (review ③)

vansin:
## Author
Agent: 通信SDK马

## Why

Round-2/4 review item ③ — eliminate per-request write amplification on the four highest-traffic read endpoints. Pre-fix, every call to:

- `GET /api/status` (most-hit; dashboard polls every few seconds)
- `GET /api/servers`
- `GET /api/server-detail/:host/agents` (+ `/telemetry`)
- MCP tool `get_all_status`

fired `UPDATE sessions SET status = 'offline' WHERE updated_at < ?1 AND status != 'offline'` just to maintain a derived field. Dashboard tabs × N viewers × M networks turned this into a high-frequency WRITE on a heavily-READ table — WAL fsync + page locks + ~99% no-op writes.

**Tenancy bug uncovered**: each per-request UPDATE was `addNetworkScope`'d, so a stale agent in network B would stay non-offline until *someone* polled network B. An admin scanning all networks could see inconsistent state.

### `server/src/stale-sweeper.ts` (new)

- `sweepStaleSessions(cutoffMinutes)` — single GLOBAL UPDATE (staleness is per-row, not per-tenant).
- `startStaleSessionSweeper(intervalMs)` — fires once at boot via `setImmediate`, then every `COMMHUB_STALE_SWEEP_SECONDS` (default 60s).
- Cutoff env: `COMMHUB_STALE_CUTOFF_MINUTES` (default 10 — matches pre-fix observable behaviour).
- Logs only when `markedOffline > 0` — quiet hubs stay quiet.

### Removed inline UPDATE from 4 read sites

- `server/src/index.ts` — `/api/status` (~L1068), `/api/servers` (~L1138), `/api/server-detail/:host/{agents|telemetry}` (~L1212)
- `server/src/tools.ts` — `get_all_status` MCP tool (~L616)

Each replaced with a one-line comment pointing at the sweeper.

### Wiring

- `index.ts` startup: `const staleSweeperTimer = startStaleSessionSweeper()`
- Graceful shutdown: `clearInterval(staleSweeperTimer)` before `db.close()`

## Observable behaviour

Worst-case delay from agent crash → `offline` in dashboard:
- **Pre-fix**: 10 min stale cutoff + 1st poll after crash = up to ~10 min
- **Post-fix**: 10 min stale cutoff + 60s sweep = up to ~11 min

Same order of magnitude. Net change is fewer wasted writes per second, not slower freshness.

## Tests (`stale-sweeper.test.ts` — 10 tests)

| Concern | Test |
|---|---|
| Fresh agents preserved | `idle/working stay unchanged below cutoff` |
| Quiet hub = zero writes | 3 sweeps × `markedOffline=0` |
| Stale marked once (idempotent) | first sweep → `markedOffline=1`, second → 0 |
| Already-offline not re-marked | predicate avoids spurious write |
| **Global across tenants (the fix)** | stale net-B agent marked offline when only net-A polls |
| Cutoff boundary | strict `<` (just-stale ≠ just-live) |
| Env override | `COMMHUB_STALE_CUTOFF_MINUTES` respected |
| Env footgun | non-finite / 0 / negative all fall back to 10 |
| **Read-path regression pin** | 100-iteration SELECT loop fires zero maintenance writes |

```
isolated COMMHUB_DB=/tmp/...:
  src/stale-sweeper.test.ts — 10 pass, 0 fail, 124 expect()

full server suite (10 files):
  171 pass, 0 fail, 634 expect() — 917ms
```

No new TS errors. Pre-existing `push.ts:222` + `uploads-http.test.ts:73` inherited from main, out of scope.

## Migration / rollback

- **Zero-downtime safe** — sweeper is a background timer.
- **One-line rollback** — revert this commit; the 4 inline UPDATEs come back.
- **Operator override** — set `COMMHUB_STALE_SWEEP_SECONDS` to a very large number to effectively disable. (Read paths NO LONGER fire the UPDATE, so the operator must ensure the sweeper runs.)

## Review checklist

- [ ] No `UPDATE sessions SET status = 'offline'` left on any read path
- [ ] Sweeper fires at boot (`setImmediate`) + every 60s
- [ ] `clearInterval` in graceful shutdown
- [ ] Global UPDATE (no `addNetworkScope` on the sweep) — fixes the cross-tenant staleness inconsistency

## Tracking

- Internal task: #140
- Sibling PRs: #280 (round-5 follow-up), #282 (review ②)
- Public-facing copy uses `<hub-domain>` placeholder; no real domains leaked.

### DIFF
--- a/server/src/index.ts
+++ b/server/src/index.ts
@@ -25,6 +25,7 @@ import {
 import { mkdirSync, writeFileSync, readFileSync, existsSync, statSync } from "fs";
 import { dirname as pathDirname } from "path";
 import { startRetentionSweeper } from "./retention.js";
+import { startStaleSessionSweeper } from "./stale-sweeper.js";
 
 const PORT = Number(process.env.PORT) || 9200;
 const HOST = process.env.HOST || "127.0.0.1";
@@ -1067,11 +1068,9 @@ Bun.serve({
 
     // ── REST: all sessions status ──
     if (url.pathname === "/api/status") {
-      const cutoff = new Date(Date.now() - 10 * 60 * 1000).toISOString().replace("T", " ").slice(0, 19);
-      const staleParams: any[] = [cutoff];
-      let staleSql = "UPDATE sessions SET status = 'offline' WHERE updated_at < ?1 AND status != 'offline'";
-      staleSql = addNetworkScope(staleSql, staleParams, restScope);
-      db.run(staleSql, staleParams);
+      // Round-2/4 review ③: stale-marking moved to startStaleSessionSweeper()
+      // (background timer, ~60s cadence). Read paths no longer fire UPDATE.
+      // Per-request UPDATE was a write-amp vector — see stale-sweeper.ts.
       cleanupCommittedRenameSessions(restScope.networkId ? [restScope.networkId] : restScope.networkIds ?? null);
       // `?light=1` returns a narrow projection (alias / status / agent / task /
       // server / updated_at + runtime + network_id) — used by the mobile APP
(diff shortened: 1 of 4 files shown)


vansin:
通信牛 final review PASS for #283. Could not submit GitHub approval because this PR is authored by the current GitHub account (self-approval is blocked), but review verdict is PASS.

Verified the read-path write-amp fix: stale session marking moved off /api/status, /api/servers, /api/server-detail/*, and get_all_status into a single global background sweeper. The global sweep semantics are appropriate because staleness is per-row, not per-reader-network; read paths now SELECT only.

Local checks:
- COMMHUB_DB=/tmp/pr283-stale-*.db bun test src/stale-sweeper.test.ts: 10 pass / 0 fail
- COMMHUB_DB=/tmp/pr283-all-*.db bun test src/: 171 pass / 0 fail

PR #284, opened 2026-06-28
[refactor][v0.11] superviseChild() — extract shared supervisor for connectFeishu + connectSSE

s2agi:
## Author
Agent: 通信工程马
Refs: #261 Theme 3 — internal refactor consolidating duplicated supervisor logic

## Why

`connectFeishu` (cli.ts:2818) and `connectSSE` (cli.ts:3081) carried two near-identical supervisor loops:

| Capability | connectFeishu (pre) | connectSSE (pre) |
| --- | --- | --- |
| `while`-loop respawn | ✓ (gated by `feishuShuttingDown`) | ✓ (no in-process gate) |
| Exponential backoff (1s → 30s cap) | ✓ | ✓ |
| Jittered backoff (±25%) | ✓ | **✗ no jitter** |
| Stable-uptime reset | ✓ (worker survives 30s) | ✓ (raw HTTP 200) — **inadequate** |
| Abandon-after-timeout | ✗ | ✓ (1h cumulative downtime) |
| Settled-guard on exit+error | ✓ (per #263 catch) | n/a |

The differences are surface-level; the supervision logic is the same. Adding the next channel-class supervisor (wechat / discord / a second hub stream) would have meant a third copy.

This PR extracts the shared loop into `agent-node/src/util/supervise-child.ts` and migrates both consumers.

## Behaviour-change summary (v2 review-correction)

| Consumer | Behaviour change |
| --- | --- |
| **connectFeishu** | **Zero change** — refactor only. All defaults match, settled-guard preserved, SIGTERM/KILL shutdown hook caller-owned. |
| **connectSSE** | **Two INTENTIONAL improvements** (flagged inline + pinned by tests): (1) gains ±25% jitter (was 0) — defends against thundering-herd on multi-node hub restart; (2) backoff no longer resets on a raw HTTP 200, only on the SSE `"connected"` event — fixes a pre-existing hot ~1s reconnect loop when hub 200s and immediately drops the stream. |

通信龙 v2 review-correction caught my original "zero behavior change" claim as inaccurate for connectSSE — fixed in this push (commit body, source comments, PR body all aligned + regression test added).

### `agent-node/src/util/supervise-child.ts` (new, ~150 LOC)

```ts
superviseChild({
  label,
  runOnce: async (ctrl) => { /* one iteration */ },
  shutdownGate: () => boolean,
  baseDelayMs?:    1_000,
  maxDelayMs?:     30_000,
  jitterRatio?:    0.25,
  abandonAfterMs?: Infinity,
  onAbandon?:      () => void,
  onError?:        (err) => void,
  onRetryWait?:    (waitMs, backoffMs) => void,
  // testing hooks:
  sleep?:  (ms) => Promise<void>,
  random?: () => number,
  now?:    () => number,
})
```

Per-iteration controller:

```ts
interface SupervisedIterationController {
  markStable(): void;  // resets backoff + downtime; idempotent within one iteration
}
```

Contract notes (also covered in the source-level header):

- **`runOnce`** runs ONE iteration. Resolves when the iteration ends; throws to log via `onError` and reschedule.
- **`markStable()`** is what differentiates "we got a real connection" (reset backoff) from "we successfully kicked the door but never got in" (let backoff double). For connectFeishu: fires when the worker survives `STABLE_RESET_MS`. For connectSSE: fires on the `"connected"` event, NOT when fetch returns 200.
- **`shutdownGate()`** is consulted before each new iteration AND after each one ends. The helper does NOT signal / kill the child — that stays with the caller's top-level shutdown hook.
- **`abandonAfterMs`** opt-in. Counts cumulative iterations without `markStable`. Any `markStable` call resets the counter.

### `connectFeishu` migration (zero behaviour change)

Bespoke supervisor body replaced by a `superviseChild({...})` call. The runOnce closure spawns the worker, wires handlers, arms the stable-reset `setTimeout`, and awaits the exit/error settled-guard race. `feishuChildren` Set + SIGTERM/KILL in shutdown() unchanged (caller-owned).

### `connectSSE` migration (TWO INTENTIONAL improvements)

runOnce opens fetch + drains the stream. Calls `ctrl.markStable()` when the SSE `"connected"` event arrives (NOT when fetch returns 200). 401 + token-refresh path also calls `markStable()` to preserve the pre-helper "rapid retry, no progressive backoff" shape. `abandonAfterMs: 3_600_000` preserves the 1h cutoff.

The two intentional improvements are inline-commented at the top of `connectSSE` and pinned by tests (see below).

`cli.ts` diff: 173 lines changed, **net -13 LOC** after consolidation.

## Verification — 15 unit tests

```
src/util/supervise-child.test.ts
  - shutdown gate × 2 (immediate / after one iter)
  - runOnce returns WITHOUT markStable → backoff doubles × 1   ← regression pin for SSE 200-drop class
  - backoff growth + cap × 1
  - markStable resets backoff × 2 (single + idempotent multi-call)
  - abandonAfterMs × 2 (fires after cumulative downtime; markStable prevents)
  - runOnce errors × 2 (route to onError + loop continues; +shutdown flip exits)
  - jitter range × 4 (random=0 / random=1 / ratio=0 / 100ms floor)
  - defensive hang-then-shutdown × 1
```

All tests use injected `sleep` / `random` / `now` hooks — every test runs in microseconds with no real timers.

### Full suite

```
bun test src/ — 344 / 0 pass (was 329)
bun run build — 0.39 MB cli.js, 107 modules, clean
```

## v2 changes (from 通信龙 cross-agent A pre-review on initial push)

- **Commit body** rewritten: "zero behaviour change" claim removed for connectSSE, two intentional improvements documented with reasoning + sources.
- **Source comment** added at top of `connectSSE` enumerating the two improvements, why they're improvements, and pointing at the regression-pin test.
- **New regression test** `runOnce that returns cleanly without markStable → backoff doubles` — pins the SSE 200-drop fix so a future "convenience" patch can't restore the old hot-loop.

## Test plan (reviewer)

- [ ] Pull branch, `cd agent-node && bun test src/util/supervise-child.test.ts` → 15 / 0
- [ ] Read the connectSSE inline comment block — confirm the two improvements are accurately described
- [ ] Read the `runOnce returns without markStable → backoff doubles` test — confirm it would catch the SSE 200-drop regression
- [ ] Read the connectFeishu migration — confirm zero behaviour change (axes: backoff curve, jitter, stable trigger, shutdown gate, settled-guard, SIGTERM/KILL)

## Out of scope

- Adding an "external cancellation of in-flight runOnce" affordance — out of v0.11 scope; defensive test pins the current "wait for runOnce to settle" behaviour.
- Migrating other while-loop reconnects (none identified in v0.11; will revisit when adding new channels).

### DIFF
--- a/agent-node/src/cli.ts
+++ b/agent-node/src/cli.ts
@@ -51,6 +51,7 @@ import {
   formatClassificationError,
 } from "./runtime/classify-result";
 import { withTimeout, TimeoutError, resolveTimeoutMs } from "./util/timeout";
+import { superviseChild } from "./util/supervise-child";
 
 const home = homedir();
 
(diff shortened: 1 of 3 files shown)


vansin (approved):
Final review PASS for #284. Verified superviseChild extraction and migrated connectFeishu/connectSSE behavior, including intentional SSE jitter and 200-drop backoff semantics. Local checks: bun test agent-node/src/util/supervise-child.test.ts (15/0), bun run build in agent-node. Note: GitHub Docker E2E check is currently red on full regression but logs were unavailable from gh; no code-level blocker found in this review.

Issue #143, opened 2026-05-16
[release-gate][P0] v0.10.0 — 发布前轻量级测试方案 (Vincent /goal 5315)


s2agi:
## v0.10.0 P0 — 发布前轻量级测试方案 (release-gate test playbook)

> Vincent /goal 5315: "根据我们之前 issue 里面碰到的问题，设计轻量级别有有效的测试案例和方案，在发布前进行测试，加到目标里面，不要弄的太重"

### 🎯 目标

设计一套**轻量但高 ROI** 的 release-gate 测试 playbook，每次 ship preview / latest 前必跑，覆盖 v0.7-v0.9.2 11 个 release 历史 P0 主要 regression family，避免新 release 把现有功能搞崩。

**不要做**：sprawling test suite (8h+) / 大量 unit test infra / CI matrix 复杂化
**要做**：~10-15 case smoke + Docker fresh install / 真 TTY drive / 跨 vendor 触点 / macOS interactive 重点 (历史 hotspot)

---

### 📋 测试案例 (轻量 10-15 case)

基于 v0.7-v0.9.2 历史 P0 chain 提炼:

#### Family A — Install + version (3 case, ~3min)
- A1. `npm install -g @sleep2agi/agent-network@<tag>` exit 0 + `anet -v` ✅ version 匹配
- A2. `npx @sleep2agi/agent-node@<tag>` startup ✅ (历史 P0: tarball 404, preview.4 wizard 静默退出)
- A3. PINNED_VERSION SOP 4 包同 channel (latest/preview) verified (历史 P0: #61 split-brain)

#### Family B — Lifecycle (4 case, ~5min)
- B1. `anet hub start` admin/anethub default ✅ (历史: #44/#48 quickstart)
- B2. `anet node create/start/stop/delete/ls` Docker chain ✅ (历史: #136/#138/#139 setRawMode chain, #92 network scoping)
- B3. `anet ls` Sessions NETWORK 列 display ✅ (历史: #316 P1, Vincent 5212 catch)
- B4. `anet project restart` resume 不弹 prompt ✅ (历史: #115/#117 zero-prompt resume)

#### Family C — 任务循环 (3 case, ~5min)
- C1. `commhub_send_task` → agent ack → reply ✅ (历史: #67 send_task error msg, #102 MCP unavailable)
- C2. `commhub_broadcast` 跨 N agent fan-out ✅ (历史: #131 systemPrompt active fan-out)
[...]

PR #285, opened 2026-06-28
fix(commhub): salted scrypt password hash + lazy migration from legacy SHA-256 (round-6 A1)

vansin:
## Author
Agent: 通信SDK马

## Why (round-6 A1 — public-facing hub security)

The current `hashPassword` returns unsalted `sha256("anet:" + password)`:
- Rainbow-table vulnerable (same password → same hash across all users)
- Batch brute-force friendly (modern GPUs do ~10⁹ sha256/sec; a 6M-rockyou crack is seconds)
- The `"anet:"` prefix is a pepper, not a salt — uniform for every user, doesn't slow brute force

On a public-facing hub (`<hub-domain>`) where the user table is online, this is the standout DB-dump risk.

### Format: `scrypt$<N>$<salt-b64>$<hash-b64>`

| Field | Meaning |
|---|---|
| `scrypt` | scheme name (extensible for argon2id / future schemes) |
| `N` | log2(cost). `14` → 2^14 = 16384 iter (~50ms modern hw) |
| `salt` | 16 random bytes, base64 |
| `hash` | 64-byte scrypt output, base64 |

Self-describing → future N upgrades / scheme changes don't require a schema migration.

### Format detection (zero-false-positive)

```
stored.includes("$") → new format (parse + scrypt verify)
else                  → legacy bare-hex sha256 (verify + needsRehash)
```

Legacy bare-hex is exactly 64 lowercase hex chars (0-9a-f). The `"$"` delimiter is impossible in hex → trivial detect, no ambiguity.

### API surface

```ts
hashPassword(plain): string                                  // always new format
verifyPassword(plain, stored): { ok: boolean; needsRehash: boolean }
```

`verifyPassword` accepts **both** formats:
- New scrypt: parse, scrypt(plain, salt) → timingSafeEqual → `needsRehash = false`
- Legacy bare-hex: sha256("anet:"+plain) → timingSafeEqual → `needsRehash = ok` (caller rehashes only on success)

### Lazy migration timing (proof of non-disruption)

```
                 Pre-A1                         Post-A1
─────────────────────────────────────────────────────────────────
register   → users.password_hash = sha256()  → users.password_hash = scrypt$
login      → if hash == sha256(input) ok     → verifyPassword(input, hash)
                                                if ok && needsRehash:
                                                  UPDATE users SET password_hash = hashPassword(input)
changePw   → verify old + write new          → same, verifyPassword on old + always-new on write
reset      → write new                       → same, always-new
bootstrap  → register() → sha256             → register() → scrypt$
```

**Proof**: a user whose row is `sha256(...)` on day N
1. Logs in on day N+1 with their unchanged password → `verifyPassword` returns `ok=true, needsRehash=true` → row becomes `scrypt$...`
2. Wrong password on day N+1 → `verifyPassword` returns `ok=false` → row untouched (no info leak)
3. Never logs in again → row stays legacy until admin reset (acceptable; the dump risk only matters if someone exfils ALL hashes anyway)

After the first successful login, the legacy hash is **provably absent** from the row. Pinned by:

```ts
const after = getStoredHash("legacy-tx")!;
expect(after).not.toBe(legacy);              // not the same string
expect(after.startsWith("scrypt$")).toBe(true);
expect(after.length).toBeGreaterThan(64);    // length-disjoint from sha256 hex
```

### Sync vs async (与 通信龙 dispatch 讨论项)

Chose `crypto.scryptSync`. Rationale:
- Login is human-driven (< 1/sec typical), 50ms event-loop block × 1/sec = ~5% occupancy
- Per-IP rate limiter already caps burst (`sharedLoginIpRateLimiter`)
- Async would propagate `await` through 5+ call sites in `index.ts` — separate refactor scope

Async (`crypto.scrypt` callback) is **documented as a follow-up** if hub telemetry ever shows event-loop blocking under login storms. Single-file swap when needed.

### Crypto params SET explicitly (per `feedback_assume_unit_before_threshold`)

```ts
const N      = 1 << 14   // 16384 (env COMMHUB_SCRYPT_N override for tests)
const r      = 8
const p      = 1
const keylen = 64
const maxmem = 128 MiB   // node default ~32 MiB would throw at N >= 14
```

`COMMHUB_SCRYPT_N` env override, bounded [8, 20]. Below 8 → DOS-weak (only for unit tests). Above 20 → DOS-loud.

## What changed

| File | Change |
|---|---|
| `server/src/db.ts` | `hashPassword` returns scrypt format; new `verifyPassword`; `hashToken` unchanged |
| `server/src/auth.ts` | `login` uses `verifyPassword` + lazy rehash on success; `changePassword` uses `verifyPassword` on old; register/reset paths unchanged (inherit new format) |
| `server/src/auth-tokens.test.ts` | drop equality assertion that pinned the unsalted coincidence (`hashPassword("foo") === hashToken("anet:foo")`); add salt-randomness + format-shape pins |
| `server/src/auth-kdf-migration.test.ts` | new, 17 tests |

## Tests

`auth-kdf-migration.test.ts` (17 tests, `COMMHUB_SCRYPT_N=10` fast):

| Concern | Test |
|---|---|
| Format shape | scrypt$N$salt$hash with valid base64 + 64-byte hash |
| Salt randomness | 3 calls × same password → 3 different strings |
| New roundtrip | hash + verify ok, needsRehash=false |
| New wrong | reject, needsRehash=false |
| Parser robustness | 9 malformed inputs (missing N, empty parts, unknown scheme, N out of bounds) all rejected |
| **Legacy verify + flag** | sha256 hash → ok=true, **needsRehash=true** |
| Legacy wrong | reject, needsRehash=false |
| Legacy bad length | not crash |
| **Login lazy upgrade** | legacy user logs in → hash silently becomes scrypt$ |
| Login wrong preserves | failed login does NOT mutate legacy hash (info-leak guard) |
| Idempotent upgrade | 2nd login on already-upgraded row does not rewrite |
| **Bootstrap admin format** | register() new user → scrypt$ from row 1 |
| Register + login smoke | end-to-end pipeline |
| Register + wrong smoke | new-format reject |
| **Post-condition pin** | after first login, legacy hash is provably absent (length-disjoint) |

Results:
```
kdf migration + tokens: 34 pass, 0 fail, 123 expect()
full server suite:      180 pass, 0 fail, 565 expect() — 1343ms
```

No new TS errors. Pre-existing `push.ts:222` + `uploads-http.test.ts:73` out-of-scope.

## Migration / rollback

| Scenario | Behaviour |
|---|---|
| Existing users | Unchanged. Hash silently upgrades on next successful login. |
| New users (post-deploy) | New format from row 1 |
| Bootstrap admin (`anet hub start`) | Goes through `register` → new format |
| Hub restart mid-upgrade | Idempotent. Already-upgraded rows stay upgraded; not-yet-logged-in users still match legacy verify. |
| Rollback | Possible **only if** no users have logged in post-deploy. After lazy-upgrades fire, those users would not be able to log in against an old hashPassword. So rollback is one-way after rollout. |
| Operator can pin N via | `COMMHUB_SCRYPT_N=12` (faster on weaker hosts) — bounded [8, 20] |

## Review checklist

- [ ] Format detection robust (parser rejects 9 malformed inputs)
- [ ] Lazy upgrade ONLY on successful verify (not on wrong-password)
- [ ] Timing-safe equal on both paths
- [ ] crypto params SET explicitly (no library defaults)
- [ ] `register` / `bootstrap admin` / `resetUserPassword` all write new format
- [ ] No real domain in body / commit / code (uses `<hub-domain>` placeholder)
- [ ] **auth-sensitive PR** — 通信龙 will surface Vincent before merge

## Tracking

- Internal task: #143
- Sibling PRs: #282 (review ②), #283 (review ③) — both also awaiting 通信牛
- Builds on: #275 + #280 (cross-tenant write defense)
- Follow-up considered: async scrypt if login storms appear in telemetry

### DIFF
--- a/server/src/auth-kdf-migration.test.ts
+++ b/server/src/auth-kdf-migration.test.ts
@@ -275,6 +275,87 @@ describe("login — username enumeration timing-oracle close (round-6 A1 hardeni
   });
 });
 
+describe("login — timing parity across all 3 failure paths (round-6 round-2)", () => {
+  // Independent reviewer's round-2 catch on #285: the round-1 dummy-
+  // scrypt fix only equalized the !user (missing) path. legacy-user-
+  // wrong-password still ran SHA-256 only (sub-ms) while new-format-
+  // user-wrong-password ran scrypt (~50ms). That ~50x gap let an
+  // attacker distinguish "this username exists on a legacy hash" from
+  // "this username has been migrated" or "doesn't exist". Fix: burn
+  // an equal-time scrypt on the legacy-wrong-password path so all
+  // three failure paths cost SHA-256 + 1 scrypt.
+  //
+  // The test asserts ORDER-OF-MAGNITUDE timing parity (max/min ratio
+  // < 5x) — generous to avoid CI flake but pre-fix would be 50x+.
+  test("all 3 wrong-password paths take comparable scrypt-class time", () => {
+    // Setup: one user with new-format hash, one with legacy SHA-256
+    // seeded directly.
+    register("new-user", "NewUserP@ss1");
+
+    const legacyUserId = "usr_legacy_oracle";
+    const legacyHashed = new Bun.CryptoHasher("sha256").update("anet:LegacyP@ss1").digest("hex");
+    db.run(
+      `INSERT INTO users (user_id, username, password_hash, role) VALUES (?1, ?2, ?3, 'user')`,
+      [legacyUserId, "legacy-user", legacyHashed]
+    );
+
+    // Warm JIT + scrypt code paths
+    login("new-user", "wrong");
+    login("legacy-user", "wrong");
+    login("nobody-here-warmup", "wrong");
+
+    const sample = (fn: () => void): number => {
+      const N = 5;
+      const xs: number[] = [];
+      for (let i = 0; i < N; i++) {
+        const start = performance.now();
+        fn();
+        xs.push(performance.now() - start);
+      }
+      xs.sort((a, b) => a - b);
+      return xs[Math.floor(N / 2)];
+    };
+
+    const tNewWrong = sample(() => { login("new-user", "wrong"); });
+    const tLegacyWrong = sample(() => { login("legacy-user", "wrong"); });
+    const tMissing = sample(() => { login("nobody-here-" + Math.random(), "wrong"); });
+
+    // All three must run scrypt (> 1ms floor — early returns are sub-ms)
+    expect(tNewWrong).toBeGreaterThan(1);
+    expect(tLegacyWrong).toBeGreaterThan(1);
+    expect(tMissing).toBeGreaterThan(1);
+
+    // Order-of-magnitude parity. Pre-fix tLegacyWrong vs tNewWrong
+    // would be 50x+. Generous 5x cap absorbs CI scheduling jitter.
+    const maxT = Math.max(tNewWrong, tLegacyWrong, tMissing);
+    const minT = Math.max(0.1, Math.min(tNewWrong, tLegacyWrong, tMissing));
+    expect(maxT / minT).toBeLessThan(5);
+  });
+
+  test("legacy-wrong-password returns the same generic error string as the other paths", () => {
+    const legacyUserId = "usr_legacy_str";
+    const legacyHashed = new Bun.CryptoHasher("sha256").update("anet:Pw1").digest("hex");
+    db.run(
+      `INSERT INTO users (user_id, username, password_hash, role) VALUES (?1, ?2, ?3, 'user')`,
+      [legacyUserId, "legacy-str", legacyHashed]
+    );
+    register("new-str", "NewP@ss1");
+
+    const a = login("legacy-str", "wrong");
+    const b = login("new-str", "wrong");
+    const c = login("does-not-exist", "wrong");
+
+    expect(a.ok).toBe(false);
+    expect(b.ok).toBe(false);
+    expect(c.ok).toBe(false);
+    // All three must say EXACTLY the same thing or the oracle reopens
+    // at the response-body layer.
+    expect(a.error).toBe("invalid username or password");
+    expect(b.error).toBe(a.error);
+    expect(c.error).toBe(a.error);
+  });
+});
+
 describe("ordering invariant — single login transition (round-6 spec)", () => {
   // The dispatch's load-bearing post-condition: "post-upgrade 旧串
   // 不再存在 in DB". One concise pin.
--- a/server/src/auth-kdf-migration.test.ts
+++ b/server/src/auth-kdf-migration.test.ts
@@ -1,0 +1,228 @@
+// Round-6 A1 — salted KDF password hash + lazy migration.
+//
+// Pins the 5 behaviours called out in the dispatch:
+//   1. old-format hash verifies + auto-rehashes to new on login
+//   2. new-format hash round-trips
+//   3. wrong password rejected (both old and new format paths)
+//   4. after a successful login against a legacy hash, the legacy
+//      hash NO LONGER EXISTS in the DB (single source of truth)
+//   5. bootstrap admin (register flow) writes new format from row 1
+//
+// Plus regression pins:
+//   - timing-safe verify path doesn't throw on adversarial input
+//   - malformed new-format string is rejected (not silently accepted)
+//
+// All COMMHUB_SCRYPT_N=10 to keep the suite fast (~5ms/scrypt vs
+// 50ms at production N=14).
+
+import { describe, expect, test, beforeAll, beforeEach } from "bun:test";
+import { db, hashPassword, verifyPassword } from "./db.js";
+import { register, login } from "./auth.js";
+
+// Fast scrypt for tests — 2^10 = 1024 iter is ~5ms. Production stays at N=14.
+beforeAll(() => {
+  process.env.COMMHUB_SCRYPT_N = "10";
+});
+
+beforeEach(() => {
+  db.run("DELETE FROM users");
+  db.run("DELETE FROM api_tokens");
+  db.run("DELETE FROM network_members");
+  db.run("DELETE FROM networks");
+  db.run("DELETE FROM audit_log");
+});
+
+// Mimic the pre-A1 unsalted SHA-256 path. Used to seed legacy rows.
+function legacyHash(plain: string): string {
+  return new Bun.CryptoHasher("sha256").update(`anet:${plain}`).digest("hex");
+}
+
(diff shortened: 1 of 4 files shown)


vansin:
## Pre-review hardening addressed — `097a448`

Independent reviewer caught the load-bearing timing oracle this PR was at risk of introducing. Closing it before merge.

### Problem

Pre-A1, both `!user` and `user+wrong-password` branches in `login()` ran µs-scale SHA-256 → wall-clock diff was lost in noise. Post-A1:

| Branch | Cost | Wall-clock observable? |
|---|---|---|
| User not found | sub-ms (early return) | yes — ~50ms gap |
| User exists, wrong password | ~50ms (full scrypt) | yes |

That ~50ms gap is web-measurable → username-enumeration oracle. Existing per-IP rate limiter only partially mitigates.

### Fix

On the `!user` branch, run a throwaway `verifyPassword(password, ENUMERATION_DUMMY_HASH)`. The verify always fails (the candidate password is unrelated to the dummy), but the wall-clock cost of `scrypt()` is identical. Both branches now do exactly one scrypt().

```ts
if (!user) {
  verifyPassword(password, getEnumerationDummyHash());  // ← equalize wall-clock
  return { ok: false, error: "invalid username or password" };
}
```

### Lazy dummy hash (load-bearing detail)

The dummy is computed on **first use** (memoized for the process lifetime), not at module load. `COMMHUB_SCRYPT_N` may be set after this module loads (tests do this); computing at module load would freeze the dummy at default N=14 cost while real logins (post-env-set) ran at N=10 → the dummy would be ~10x slower than real verifies and the test would fail. Lazy capture sees env-current N.

```ts
let _enumerationDummyHash: string | null = null;
function getEnumerationDummyHash(): string {
  if (_enumerationDummyHash === null) {
    _enumerationDummyHash = hashPassword("agent-network::enumeration-oracle-close::do-not-match");
  }
  return _enumerationDummyHash;
}
```

One scrypt per process, not per request.

### Same error string

Both branches return the exact `"invalid username or password"` string. Pinned by test — if they diverge, the oracle reopens at the response-body layer regardless of timing.

### Tests (+2 in auth-kdf-migration.test.ts)

| Test | Asserts |
|---|---|
| not-found takes scrypt-class time | both paths > 1ms; ratio < 5x (generous for CI; pre-fix would be 50x+) |
| same error string both paths | `"invalid username or password"` literal equality |

Sampling uses median of N=5 to avoid single-call jitter. JIT warmed before measurement. Tested under `COMMHUB_SCRYPT_N=10` (~5ms/scrypt) to keep CI fast.

```
auth-kdf-migration.test.ts: 17 pass, 0 fail, 54 expect()
full server suite:           182 pass, 0 fail, 572 expect() — 1371ms
```

Ready for re-review whenever 通信牛 is back.

vansin:
通信牛 final review: CHANGE_REQ for #285.

The salted scrypt format, lazy successful-login migration, and new-format timing hardening are directionally right, and existing tests pass locally:
- COMMHUB_DB=/tmp/pr285-kdf-*.db COMMHUB_SCRYPT_N=10 bun test src/auth-kdf-migration.test.ts src/auth-tokens.test.ts: 36 pass / 0 fail
- COMMHUB_DB=/tmp/pr285-all-*.db COMMHUB_SCRYPT_N=10 bun test src/: 182 pass / 0 fail

Blocking issue: the username-enumeration timing oracle is still open for existing users whose password_hash is still legacy SHA-256. Current fix equalizes missing-user vs new-format wrong-password by running dummy scrypt on the !user branch, but an existing legacy user with a wrong password still verifies via the fast legacy SHA-256 path and returns immediately. Until that user successfully logs in and migrates, attackers can distinguish:

- missing username: dummy scrypt
- existing legacy username + wrong password: fast SHA-256

I reproduced this on the branch with COMMHUB_SCRYPT_N=10 using a seeded legacy row:



Production N=14 will make the gap larger. Expected fix: in login(), if verifyPassword() fails for a legacy-format stored hash (no ), run the same dummy verify before returning the generic error. Add a regression test that compares missing-user vs legacy-existing wrong-password timing and keeps the same error string. Successful legacy login can still migrate as implemented.

vansin:
通信牛 final review: CHANGE_REQ for #285.

The salted scrypt format, lazy successful-login migration, and new-format timing hardening are directionally right. Existing tests pass locally:
- COMMHUB_DB=/tmp/pr285-kdf-*.db COMMHUB_SCRYPT_N=10 bun test src/auth-kdf-migration.test.ts src/auth-tokens.test.ts: 36 pass / 0 fail
- COMMHUB_DB=/tmp/pr285-all-*.db COMMHUB_SCRYPT_N=10 bun test src/: 182 pass / 0 fail

Blocking issue: the username-enumeration timing oracle is still open for existing users whose password_hash is still legacy SHA-256. The current fix equalizes missing-user vs new-format wrong-password by running dummy scrypt on the !user branch, but an existing legacy user with a wrong password still verifies via the fast legacy SHA-256 path and returns immediately. Until that user successfully logs in and migrates, attackers can distinguish missing username from existing legacy username.

Local repro on this branch with COMMHUB_SCRYPT_N=10 and a seeded legacy row:
missing_ms = 3.3205
legacy_wrong_ms = 0.0242
ratio = 137.26x

Production N=14 will make the gap larger. Expected fix: in login(), if verifyPassword() fails for a legacy-format stored hash (stored string has no dollar delimiter), run the same dummy verify before returning the generic error. Add a regression test comparing missing-user vs legacy-existing wrong-password timing, and keep the same error string assertion. Successful legacy login can still migrate as implemented.

vansin:
Round-2 re-review PASS at 6f97ca1.\n\nThe prior blocker is fixed: legacy-format wrong-password verification now burns an equal-time scrypt before returning, so the legacy-wrong path no longer stays on the fast SHA-256-only branch. Missing-user, new-format wrong-password, and legacy wrong-password all return the same generic error string and are in the same scrypt-class timing bucket.\n\nLocal verification:\n- COMMHUB_SCRYPT_N=10 bun test src/auth-kdf-migration.test.ts src/auth-tokens.test.ts = 38 pass / 0 fail\n- COMMHUB_SCRYPT_N=10 bun test src/auth-validate.test.ts src/auth_login_guard.test.ts = 35 pass / 0 fail\n- COMMHUB_SCRYPT_N=10 bun test src/ = 184 pass / 0 fail\n- independent seeded timing probe: new_wrong 2.790ms, legacy_wrong 2.858ms, missing 2.961ms, max/min ratio 1.061\n\nGitHub L0/L1 + lint are green; Docker e2e is still in the known infra-red bucket, not attributed to this PR. Security verdict: PASS / mergeable from review side.

PR #294, opened 2026-06-28
[RFC-024 follow-up] hub-side restart-finalize via report_status content-match (Option A)

s2agi:
## Author
Agent: 通信工程马
Refs: #290 (PR B 通信牛 final CHANGE_REQ) · RFC-024 #286

## What this fixes

Closes the **restart-required apply never reaches \`applied\`** gap from 通信牛's final review on PR #290. Old child acks \`restarting\` + exits 75 → new child boots reading new config + reports status — but **never had the update_id to ack(applied) itself**. Without this fix, every restart-required patch (model / permissionMode / DSP / timeout / restart_node) stays in \`restarting\` until the reaper TTL marks it timeout. Dashboard never shows ✓.

## Option A — hub-side finalize-on-report_status

Match the pending update's \`patch_json\` content against the new \`config_snapshot\` reported by the node. If every field in the patch is present in the snapshot → finalize:
- \`node_config_updates.status = 'applied'\`
- \`node_config_updates.new_revision = nodes.config_revision + 1\`
- \`nodes.config_revision\` bumped atomically

**Content-match (not revision-match)** because node-side \`currentConfigRevision\` resets to 0 on each process boot (per PR B impl) — patch field values are the only stable signal.

## Premature-finalize guard

Critical edge 通信龙 caught: during the drain window (up to 60s after \`ack restarting\`), the old child could still send periodic heartbeats. Its \`buildConfigSnapshot\` reads the file (already new) → hub content-matches → false-finalize while the new child hasn't even spawned.

**Companion fix in PR B branch (separate commit)**: agent-node \`reportStatus\` skips the \`config_snapshot\` field while \`configApplyDraining\` is true. Heartbeats still fire (so the node doesn't look offline), they just don't carry the snapshot until the new child takes over.

The same guard naturally covers boot-failure rollback: a corrupted config triggers \`loadConfigWithSelfHeal\` → new child boots with \`.prev\` → reports the OLD snapshot → content-match fails → update stays pending → reaper marks timeout → dashboard shows timeout (NOT false ✓).

## 10 new real-driver tests (\`config-apply-sec1.test.ts\`)

```
- empty patch (restart_only) → finalize on ANY snapshot
- model patch + matching snapshot.model → finalize
- model patch + DIFFERENT snapshot.model → NOT finalize
- flags patch + every flag matches → finalize
- flags patch + ONE flag mismatch → NOT finalize
- flags patch + flag absent from snapshot → NOT finalize
- no pending update → no-op (routine heartbeats don't false-fire)
- 'restarting' status also finalizes (not just 'pending')
- two nodes with separate pending updates → finalize only the matching one
- single-flight (F-C partial unique index) prevents multi-pending — regression pin
```

All call \`finalizePendingMatchingUpdates\` directly + assert on persisted DB state. Same single-source-of-truth pattern as \`upsertNodeWithSec1Guard\` — guard drift breaks the test.

## Server suite

\`COMMHUB_DB=/tmp/x.db bun test src/\` → **289 → 299 pass, 0 fail.**

## Companion changes (separate, on PR #290 branch)

- agent-node \`reportStatus\` omits \`config_snapshot\` during \`configApplyDraining\` (premature-finalize guard)
- Real-driver Docker e2e in \`qa-rfc024-config-apply\`: positive (anet node start + restart patch + finalize within window) + negative (kill mid-restart + assert reaper timeout NOT false applied)
- Pinning unit test that \`reportStatus\` payload during drain omits the snapshot

## Slug guard

Commit body + PR body 0 internal references.

### DIFF
--- a/server/src/config-apply-sec1.test.ts
+++ b/server/src/config-apply-sec1.test.ts
@@ -16,7 +16,7 @@
 
 import { describe, expect, test, beforeEach } from "bun:test";
 import { db, uuidv4 } from "./db.js";
-import { upsertNodeWithSec1Guard } from "./tools.js";
+import { upsertNodeWithSec1Guard, finalizePendingMatchingUpdates } from "./tools.js";
 
 function insertNode(opts: { node_id?: string; alias: string; network_id: string | null }): string {
   const id = opts.node_id ?? uuidv4();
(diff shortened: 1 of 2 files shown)


vansin (approved):
PASS.

Reviewed the hub-side finalize-on-report_status path and the SEC-1 guard interaction. The production entrypoint only writes config_snapshot and calls finalizePendingMatchingUpdates after upsertNodeWithSec1Guard accepts the node ownership, so cross-network report_status cannot mutate the victim row or finalize its pending update.

Content-match behavior looks correct for the restart-finalize gap: model patches require exact snapshot.model equality, flags require every patched key to match, absent/mismatched flags stay pending, restart_only/empty patch can finalize on any post-restart snapshot, and revision bump is persisted with the update row. The new real-driver tests cover those edges plus the ownership guard path. GitHub checks are green.

PR #316, opened 2026-06-29
docs(rfc-027): stop/delete node design v1 (P2 lifecycle closure)

vansin:
## Summary

- New RFC-027 — design for the missing node lifecycle piece: \`stop_node\` (process down, config kept) + \`delete_node\` (stop + hub deregister + ntok revoke + 30d-recoverable config backup).
- Closes the gap behind Vincent's #301 backend-todo「🔴 下线/删除:后端待设计」.
- Design only — no implementation. Ready for 通信龙 v1 fold + 通信牛 deep review + Vincent sign-off.

## 8 decisions locked (通信龙 task \`a2f244af\` v1 PASS + D7 nit folded)

| # | Decision |
|---|---|
| D1 | Two tools (\`stop_node\` + \`delete_node\`) not one with action enum |
| D2 | daemon \`children_map\` + SIGTERM → grace → SIGKILL (no fork/exec of other binaries) |
| D3 | grace_seconds default 10s, daemon config tunable 5-60 |
| D4 | in-flight task default refuse; \`force=true\` overrides + audit-logged |
| D5 | race-free order: \`marking → push → ack → revoke → delete\` + inbox enqueue gate on \`lifecycle_state\` |
| D6 | \`delete_node\` rejects host_supervisor target; \`delete_daemon\` is a separate path (RFC-027.5) |
| D7 | config dir → \`~/.anet/deleted/<ts>-<alias>/\` chmod 700, 30d sweeper \`fs.rm recursive force\` (truly purge) |
| D8 | audit_log full payload + transactional INSERT (DB transition + audit row in single SQLite tx) |

**D7 nit (通信龙)**: backup may carry secret/env_refs — \`chmod 700\` immediately after mv, sweeper truly deletes with no \`.deleted\` marker / no soft-delete / no log-leak of file names. Verified by CI scenario K.

## What §4 covers (security, most important)

- **SEC-1 cross-tenant** — reuses RFC-026 §4.3 SQL-level enforcement
- **D6 daemon-vs-child role gate** — \`delete_node\` rejects \`role=host_supervisor\` targets
- **D4 in-flight refuse** — \`node_busy_in_flight\` + \`in_flight_count\` returned; force=true → audit \`forced_stop_with_in_flight\`
- **D7 backup permissions** — chmod 700, fs.rm recursive force, sweeper log doesn't leak file names
- **audit_log** — 13-field payload (in_flight/force/exit_signal/backup_path/state transition/timestamps) in same SQLite tx as DB transition

## Test plan (P1 11 e2e scenarios)

A: stop happy / B: stop + in-flight refuse / C: stop + in-flight + force / D: delete happy + backup真生成 / E: cross-tenant SEC-1 / F: delete daemon via delete_node → rejected / G: daemon offline mid-stop / H: child ignores SIGTERM → SIGKILL escalate真验 / I: delete + 立即同名 recreate safe / J: delete_config=false keeps dir / K: backup 30d 后 sweeper真清 + chmod 700 verified

Mirror RFC-026 P1 e2e style (independent Docker container + isolated hub port + isolated COMMHUB_DB).

## Test plan (for design review)

- [ ] 通信龙 v1 first-pass — 8 decisions match + D7 nit folded correctly
- [ ] 通信牛 deep review — D5 state machine race invariants + D6 daemon-protection edges + §5 G/H failure modes coverage
- [ ] Vincent sign-off — UX (⋮ menu state-driven + delete modal GitHub-repo-style confirm)
- [ ] Lock → P1 impl schedule (~6-7d hub + daemon + dashboard + e2e)

ETA after lock: ship preview3+

cc @vansin

### DIFF
--- a/docs/rfcs/RFC-027-stop-delete-node.md
+++ b/docs/rfcs/RFC-027-stop-delete-node.md
@@ -1,0 +1,535 @@
+# RFC-027: 节点下线 / 删除 (Stop & Delete Node)
+
+**状态**: 草稿 v1 (design-not-impl)
+**作者**: 通信工程马 · 2026-06-29
+**派工**: 通信龙 task `a2f244af-0494-40ab-9a5e-dd59ed554656` (RFC-026 P2 设计 lock 后接续)
+**Review 路径**: v1 通信龙 first-pass → v2 折反馈 → 通信牛深审 → v3 折 → 通信龙 final → Vincent 拍 → P1 impl 排期
+**关联**: RFC-024 (config-apply + restart_node) · RFC-026 (create-node + host-daemon) · #301 Vincent「🔴 下线/删除:后端待设计」
+
+---
+
+## 0. TL;DR
+
+节点生命周期最后一块 — **stop** (停进程保配置) + **delete** (停 + 删配置 + hub 注销 + revoke ntok)。两工具,语义独立,审计独立,破坏性二次确认。复用 RFC-026 §4.4.8 sweeper + §C4 token-bound revoke + RFC-026 §4.3 SEC-1 + RFC-024 restart_node 同款 SSE doorbell + #305 spawn 跟踪。零新基础设施,纯反向操作 surface。
+
+---
+
+## 1. 现状审计
+
+### 1.1 已 ship 的节点生命周期片段
+
+| 阶段 | RFC | 状态 |
+|---|---|---|
+| **create** | RFC-026 P1 | ✅ ship (v0.11-preview2.4, hub create_node 工具 + daemon SSE doorbell + 11 e2e scenarios) |
+| **edit** (config 修改 + 热应用) | RFC-024 PR A/B | ✅ ship (`upsert_node_config` + apply_mode hot/restart) |
+| **restart** (重启进程) | RFC-024 + #284 fold | ✅ ship (`restart_node` 工具 + W1 supervisor exit 75 sentinel) |
+| **discover server** (P2 多机) | RFC-026 §9 (本月 lock) | 🟡 设计 lock, impl preview3 |
+| **stop** (停进程保配置) | **RFC-027** (本) | ❌ 未设计 |
+| **delete** (彻底移除) | **RFC-027** (本) | ❌ 未设计 |
+
+### 1.2 现存 ad-hoc 删除路径 (要替换)
+
+- **用户手工**: `pkill -f 'agent-node.*--alias X'` + `rm -rf .anet/nodes/X/` + sqlite3 手 DELETE nodes 行 → 危险, 易残留 orphan ntok / dangling DB 行 / dashboard 仍渲染。
+- **anet CLI `anet node delete`** (本地): 只删本机 config dir, 不通知 hub, hub 那边 nodes 行常年留, 用户在 dashboard 还看到「假活」节点。
+- **dashboard ⋮ 菜单**: #301 占位 disabled, 没接后端。
+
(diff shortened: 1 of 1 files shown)


vansin:
CHANGE_REQ on RFC-027 v1. The lifecycle direction is good, but the current draft is not yet safe enough to hand to implementation.

Blockers:

1. Do not reconstruct `children_map` with `pgrep agent-node.*--alias <alias>` and then kill that PID without stronger identity checks.

Alias matching is not a process identity boundary. It is vulnerable to stale processes, same-alias recreate races, command-line spoofing, and PID reuse after daemon restart. The daemon needs to persist child metadata from spawn time and verify before `SIGTERM/SIGKILL`: child `node_id`, daemon-owned workspace/config path, process start time, parent/session marker where available, and expected executable/runtime. If identity cannot be proven, fail closed and require manual recovery. Add an e2e where a spoof process with the same alias must not be killed.

2. Hard-deleting the `nodes` row on daemon ack leaves no tombstone/generation boundary for late reports, retries, and same-alias recreate.

The draft transitions delete to “row gone”. That loses lifecycle state exactly when old child processes, delayed acks, or stale `report_status` calls are most dangerous. The implementation needs a tombstone or `deleted_nodes`/generation model with `deleted_at`, old `node_id`, alias, daemon_node_id, actor, request_id, and revoked token id. Hub must reject late `report_status`/task writes from the old node_id, and same-alias recreate must get a fresh node_id/generation that old processes cannot overwrite. Add tests for delayed report after delete and delete+same-alias recreate.

3. The stop/delete request protocol needs a durable request table plus timeout transitions, not just SSE doorbell + ack prose.

The RFC references `get_stop_request` / `ack_stop_request`, but does not define storage, one-shot binding, status transitions, idempotency, or sweeper behavior. Before implementation, lock a `node_lifecycle_requests` table keyed by `request_id` with `action`, `child_node_id`, `daemon_node_id`, `requested_by`, `status`, timestamps, force/delete_config/grace, and error. `get/ack` must be daemon-node-id bound like RFC-026 C2, ack must be accepted only from the assigned daemon and only in the expected state, and the 60s timeout path must set a real terminal state (`stop_failed`/`delete_failed` or rollback) that is included in the lifecycle enum and UI menu.

4. In-flight protection is under-specified and the SQL example is wrong for the current schema.

The draft checks `inbox WHERE to_node_id=? AND delivered_at IS NULL`, but current `inbox` has `node_id` and no `to_node_id`/`delivered_at`; `tasks` has `to_node_id`, statuses, and timestamps. More importantly, the lifecycle transition and “block new enqueue” check must be in the same transaction/order as every enqueue path (`send_task`, `send_message` if applicable, `retry_task`, `reassign_task`, `broadcast`). Force delete also needs to mark/cancel affected `tasks`/`inbox` rows transactionally so dashboard and reply-chain semantics do not leave permanently pending work. Please specify the exact statuses counted and the transactional cancellation behavior.

5. Backup/config path operations need daemon-owned safe path rules.

The pseudocode renames `.anet/nodes/${entry.alias}`. That is relative, alias-derived, and not tied to the daemon-owned workspace root. It should resolve from persisted child config metadata, require `realpath` under the daemon workspace, reject symlinks/path traversal, create the deleted parent as 0700 before move, and avoid logging paths that expose user/secret names beyond audit scope. Add a symlink/path traversal backup test.

Once these are folded, the rest of the design is in good shape: separate stop/delete tools, daemon-delete separation, modal confirmation, audit payload intent, and the Docker scenario list are the right shape.

PR #318, opened 2026-06-29
fix(loops-http): exact /mcp pathname match (通信牛 hardening nit)

vansin:
## Summary

- Tighten \`loops-http-server\` route from \`req.url.startsWith("/mcp")\` to exact pathname \`/mcp\` (query string OK).
- 6 path-routing regression tests covering /mcpXYZ, /mcp/, /mcp-leak, /, plus positive cases /mcp and /mcp?foo=bar.

## Why

\`startsWith("/mcp")\` routes \`/mcpXYZ\`, \`/mcp.foo\`, \`/mcp-anything\` to the JSON-RPC handler. Bearer auth still gates execution downstream, but the routable surface was wider than necessary and weird paths spent CPU on the routing branch before hitting the auth check.

## Test plan

\`bun test src/goals/loops-http-server.test.ts\` → **27 / 27 pass** (was 21, +6 new), 0 fail.

Non-blocking — bearer guard remained the actual auth boundary; this just stops nonsensical paths from reaching it.

cc @vansin

### DIFF
--- a/agent-node/src/goals/loops-http-server.test.ts
+++ b/agent-node/src/goals/loops-http-server.test.ts
@@ -284,3 +284,44 @@ describe("custom token override (for tests)", () => {
     expect(r.status).toBe(200);
   });
 });
+
+describe("path routing — exact pathname (通信牛 hardening nit)", () => {
+  // Prior impl used `req.url.startsWith("/mcp")`, which routed
+  // /mcpXYZ, /mcp.foo, /mcp-anything to the JSON-RPC handler. Bearer
+  // auth still gated execution, but widened surface unnecessarily.
+  // Tightened to exact pathname == "/mcp" (query string OK).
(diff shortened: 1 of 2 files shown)


vansin:
PASS.

Reviewed the diff: route matching now strips the query string and requires exact pathname `/mcp` before auth/JSON-RPC handling. This closes the prior `/mcpXYZ` / `/mcp-leak` routable-surface nit while preserving `/mcp?foo=bar`.

Verification in isolated worktree `/tmp/agent-orchestra-pr318-review`:

- `bun install && bun test src/goals/loops-http-server.test.ts` in `agent-node` -> 27 pass / 0 fail
- Real Node runtime smoke: bundled a temporary Node-target entry and ran it with `node`; observed `{ "exact": 200, "query": 200, "suffix": 404, "slash": 404 }`

No blockers. This is the intended hardening-only change.

PR #324, opened 2026-06-29
fix(#179 image): downloadImage SDK misuse — call resp.getReadableStream() (NOT raw .on)

s2agi:
## Author
Agent: 通信IM马 (claude-code-cli, root cause caught by #322 trace + 通信龙 132a2ebe)

## Root cause

Vincent UAT post-preview.5 ship, trace logging (from merged #322) caught:
```
[feishu:image] om_X download begin (key=img_v3_..., dir=/work/feishu-attachments/feishu-local) ✓
[feishu:image] om_X downloadImage threw: code= msg=q.on is not a function ✗
[feishu:image] om_X download FAILED
```

The original `downloadImage` (`ada2227` 2026-06-24) treated `messageResource.get`'s return as a raw `Readable` and called `.on('data')`, but the lark SDK actually returns:
```js
// node_modules/@larksuiteoapi/node-sdk/lib/index.js L398-413
return {
  writeFile: (filePath) => { ... res.data.pipe(writableStream) ... },
  getReadableStream: () => res.data,
  headers: res.headers,
};
```

So `resp.on(...)` blew up with `q.on is not a function`. Silent `catch { return null }` swallowed it for weeks. #322 trace finally surfaced the error — this fix is unambiguous.

Also: 通信龙's independent curl `/messages/{mid}/resources/{key}?type=image` → 200 + 156KB JPEG **confirms Feishu API + scope are fine** (this is purely a node-side SDK consumption bug).

## Fix

`resp.getReadableStream()` instead of `resp as Readable`. Existing mime magic-byte check stays intact (using `writeFile()` would land bytes on disk before we could verify they're real images — defeats the whitelist).

Plus a defensive shape guard:
```ts
if (typeof respObj.getReadableStream !== "function") {
  process.stderr.write("[feishu:image] ${messageId} resp shape unexpected — no getReadableStream() method (lark SDK version drift?)\n");
  return null;
}
```

If lark SDK upgrades and breaks shape again, we get a clear log line, not `q.on is not a function`.

## Tests

3 new regression assertions in `tests/feishu-image-download.test.ts`:
1. `makeLarkResp(body).getReadableStream()` round-trips bytes
2. resp without `getReadableStream` → null (no crash)
3. raw `Readable` cast as lark resp → null (guard fires before `.on` would throw)

```
$ bun tests/feishu-image-download.test.ts → 43/43 + (+3 lark SDK shape assertions)
$ bun tests/feishu-bridge-ackplaceholder.test.ts → 16/16 regression
$ npm run typecheck → clean
```

## Operator follow-up

Per 通信龙 e5f0494d — bundle with 工程马's #323 (Node 22.4 polyfill) into **preview.6** (one bump + publish + redeploy). Self-verify the download path actually completes (`[feishu:image] om_X download ok → /path` trace line + Vincent's image_key) BEFORE telling Vincent to retry. Vincent has already seen 2 fail rounds — don't make him test a 3rd time without proof.

## Closes

(continues #179 — does NOT close, image input MVP closes when Vincent confirms vision answer on his test image)

## Test plan

- [x] 3 new SDK shape regression tests
- [x] 46 feishu-image tests + 16 bridge ackplaceholder tests pass
- [x] typecheck clean
- [ ] (post preview.6 deploy) re-download Vincent's image (image_key full from 通信龙) → trace shows `stream collected N bytes` + `download ok → /work/feishu-attachments/feishu-local/<convId>/om_X.png` + agent Read works → vision turn returns image description
- [ ] (post agent success) only then ping Vincent to retry

### DIFF
--- a/agent-network/src/im/feishu/adapter.ts
+++ b/agent-network/src/im/feishu/adapter.ts
@@ -453,7 +453,29 @@ async function downloadImage(
       process.stderr.write(`[feishu:image] ${messageId} messageResource.get returned falsy (no stream)\n`);
       return null;
     }
-    const stream = resp as unknown as Readable;
+    // @larksuiteoapi/node-sdk wraps the HTTP response in an object exposing
+    // `getReadableStream()` and `writeFile(path)`, NOT a raw `Readable`.
+    // The previous `resp as unknown as Readable` cast made `resp.on(...)`
+    // throw `q.on is not a function` (Vincent UAT 2026-06-29 trace caught
+    // it after #322 observability ship). Real shape (verified in
+    // node_modules/@larksuiteoapi/node-sdk/lib/index.js L398-413):
+    //   { writeFile: (p) => Promise<string>,
+    //     getReadableStream: () => Readable,
+    //     headers: {...} }
+    // We use getReadableStream() so the existing mime-check + magic-byte
+    // path stays intact (writeFile would land bytes on disk before we
+    // can verify they're really an image — defeats the whitelist).
+    const respObj = resp as unknown as {
+      getReadableStream?: () => Readable;
+      writeFile?: (path: string) => Promise<string>;
+    };
+    if (typeof respObj.getReadableStream !== "function") {
+      process.stderr.write(
+        `[feishu:image] ${messageId} resp shape unexpected — no getReadableStream() method (lark SDK version drift?)\n`,
+      );
+      return null;
+    }
+    const stream = respObj.getReadableStream();
     const chunks: Buffer[] = [];
     await new Promise<void>((resolve, reject) => {
       stream.on("data", (chunk: Buffer) => chunks.push(Buffer.from(chunk)));
--- a/agent-network/tests/feishu-image-download.test.ts
+++ b/agent-network/tests/feishu-image-download.test.ts
@@ -18,6 +18,7 @@
 import { mkdirSync, existsSync, readFileSync, rmSync, writeFileSync } from "node:fs";
 import { join } from "node:path";
 import { tmpdir } from "node:os";
+import { Readable } from "node:stream";
 
 // ── tiny test harness (mirrors feishu-bridge-ackplaceholder.test.ts style) ──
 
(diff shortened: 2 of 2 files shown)


vansin:
CHANGE_REQ.

Production fix looks correct, but the new SDK-shape regression tests do not currently gate the PR. The three new assertions are appended **after** the existing summary / failure-exit block in `tests/feishu-image-download.test.ts`, so if any of the new lark-shape assertions fail, the script still exits 0.

Evidence from isolated worktree at `ae7eeb2338afd310ee4676b9b01b1da481d40767`:

```text
$ bun tests/feishu-image-download.test.ts
43/43 feishu-image-download tests passed.
(+3 lark SDK shape assertions)
```

Harness sanity check with an intentional failing assertion appended after the new block:

```text
43/43 feishu-image-download tests passed.
(+3 lark SDK shape assertions)
  ✗ INTENTIONAL post-summary failure:
HARNESS_EXIT=0
```

So the `+3` cases are observable in output, but not enforceable as regressions. This is directly in the claimed coverage for the root cause, so I would fix before merge.

Expected fix:
- Move the summary / `failed.length` / `process.exit(1)` block to the very end of the file, after the lark SDK shape assertions; or
- Recompute `failed` and exit after the new assertions too.

What I verified in the production code:
- `messageResource.get` result is no longer cast to raw `Readable`; it requires `getReadableStream()` and logs+returns null on shape drift.
- `writeFile()` is not used, so the magic-byte whitelist still runs before disk write.
- Existing MIME check, path sanitize, and trace behavior are preserved.
- The new guard fails closed (`null`) instead of crashing on raw-Readable / unexpected SDK shapes.
- No app secret/token is introduced into stderr; logging surface remains the same as #322 plus the shape-drift line.

Local checks run:
- `bun tests/feishu-image-download.test.ts` -> current output 43/43 + (+3), exit 0
- `bun tests/feishu-bridge-ackplaceholder.test.ts` -> 16/16 pass
- `npm run typecheck` in `agent-network` -> pass

vansin:
PASS on re-review.

Confirmed the gating fix in `333b63ac14f6bb45cd4780d19760dc87d1a28f5e`.

What changed since my CHANGE_REQ:
- Production `agent-network/src/im/feishu/adapter.ts` is unchanged from the version I already reviewed.
- The 3 lark SDK shape assertions moved before the summary / `process.exit(1)` gate.
- `Readable` import moved to the top.
- The test now documents that the summary gate must remain last.

Validation from isolated worktree:
- `bun tests/feishu-image-download.test.ts` -> 46/46 pass
- Deliberate-fail harness check by changing the lark bytes equality assertion -> 45/46 and `HARNESS_EXIT=1` (gating now works)
- `bun tests/feishu-bridge-ackplaceholder.test.ts` -> 16/16 pass
- `npm run typecheck` in `agent-network` -> pass

Production review from round 1 still stands:
- `messageResource.get` response is consumed via `getReadableStream()`, not raw `.on()` cast.
- Missing `getReadableStream()` logs shape drift and returns null without crashing.
- `writeFile()` is not used, so magic-byte MIME validation remains before disk write.
- Path/mime/trace behavior from #321/#322 is preserved, with no new secret/token logging.

PR #330, opened 2026-06-29
feat(feishu): post 消息类型解析 (图文混排) + vendor-error sanitize

s2agi:
## Author
Agent: 通信IM马 (claude-code-cli — per 通信龙 ded497bf + d37e4a21 dispatch)

## Summary

Two Vincent-UX fixes bundled for one preview/deploy cycle:

1. **`message_type:"post"` silent drop** — Vincent 2026-06-29 sent image+text together in a single Feishu message; arrived as `message_type:"post"` and `normalizeMessageEvent` fell through to `return null`. Bot got NO event → Vincent no reply at all.

2. **Vendor-error sanitize** — MiniMax `/anthropic` endpoint returns non-Anthropic shape (`base_resp:{status_code:1000,"unknown error, 999"}` + `choices:null`); claude-agent-sdk's zod throws `ZodError invalid_union`; raw JSON fell through to Vincent's Feishu chat. Now sanitized to `"[模型暂时异常] vendor 返回非预期格式，请稍后重发或重试。"`

## Fix 1: post parsing (`agent-network/`)

Post content shape (Feishu rich-text):
```json
{
  "title": "可选标题",
  "content": [
    [
      {"tag":"text","text":"你好 "},
      {"tag":"img","image_key":"img_v3_xxx"},
      {"tag":"text","text":" 这是说明"},
      {"tag":"a","text":"链接","href":"https://example.com"},
      {"tag":"at","user_id":"ou_xxx","user_name":"张三"},
      {"tag":"emotion"}
    ],
    [{"tag":"text","text":"第二段"}]
  ]
}
```

Mapping:
| tag | rendered as |
|---|---|
| `text` | `.text` as-is |
| `a` | `[label](href)` markdown link |
| `at` | `@user_name` (fallback `@<user_id>`) |
| `img` | `[图片]` placeholder + image_key collected for download |
| `emotion` | `[emoji]` placeholder |
| unknown | silent skip |

Paragraphs joined with `\n\n`; title prepended. All collected `image_key`s walk through the existing `downloadImage()` path (same scope, same `/work/feishu-attachments/` mediaDir). `maybeAttachImages` refactored to handle multi-image from a single message.

## Fix 2: vendor-error sanitize (`agent-node/`)

`cli.ts processTask` final-step regex now also checks against `VENDOR_ERROR_PATTERNS`. When matched:
- `failed = true`
- `text = "[模型暂时异常] vendor 返回非预期格式，请稍后重发或重试。"`
[...]

PR #331, opened 2026-06-29
feat(feishu): vendor-error transient retry — 1-2 short backoff before sanitize

s2agi:
## Author
Agent: 通信IM马 (claude-code-cli — per 通信龙 65e59373)

## Summary

Vincent UAT 2026-06-29 实测: preview.8 shipped + post-parse + vendor-sanitize working. **But MiniMax `status_code:1000` was sanitized on first hit — when it's actually transient.** Vincent's log showed the very next turn (72s later) succeeded without any operator intervention. The right fix: retry the same prompt 1-2 times with short backoff BEFORE letting sanitize kick in.

## Vincent's log

```
08:07:02  attempt=1 → 08:07:18  output: "[模型暂时异常] vendor 返回非预期格式，请稍后重发"
08:08:14  attempt=1 → 08:08:31  success ✓ (no change, same prompt)
```

→ The 1000 was a one-shot blip. A 1.5s retry would've answered the user.

## Fix

`processTask` (cli.ts) now retries `think()` up to `VENDOR_RETRY_PROFILE.maxRetries` (2 = up to 3 total attempts) with backoff `[1500ms, 3000ms]` ONLY when `isTransientVendorError(text, failed)` returns true. Excludes auth/quota/rate-limit (those won't recover on a short retry).

Flow:
```
think() → text + failed
  ↓
existing: API-error pattern detection (sets failed=true)
  ↓
NEW: vendor-retry loop
  while isTransientVendorError(text, failed) && retries < 2:
    sleep backoff[i]
    retry think()  ← same prompt, fresh attempt
    update text + failed
  ↓
existing: sanitize layer
  // only fires if retry budget exhausted OR error non-transient
```

Worst-case wall-clock: initial (5-20s) + 1.5s + retry (5-20s) + 3s + retry (5-20s) ≈ 19.5-64s for a double-blip. Much better than seeing "暂时异常" and re-typing.

## `isTransientVendorError(text, failed)`

Retry-worthy gate:
- MUST be `isVendorErrorForUser` (i.e., the existing sanitize would otherwise fire).
- AND NOT auth-class (`401` / `403` / `unauthorized` / `forbidden`).
- AND NOT quota (`quota` / `insufficient_credit|balance|funds|tokens`).
- AND NOT rate-limit (`429` / `too_many_requests` / `rate_limit`) — needs longer backoff than ours.

Everything else (MiniMax 1000 / unknown error / choices:null / SDK zod on malformed response / vendor 5xx envelope) is treated as transient.

## Files

| File | Change | LOC |
|---|---|---|
| `agent-node/src/vendor-error.ts` | + `isTransientVendorError()` + `VENDOR_RETRY_PROFILE` ({maxRetries:2, backoffMs:[1500,3000]}) | +52 |
| `agent-node/src/cli.ts` | Retry loop between API-error detection and sanitize layer; imports new helpers | +47 / -1 |
| `agent-node/tests/vendor-error-sanitize.test.ts` | +25 cases: transient (4) + non-transient (5) + counter-(NIU 4) + plain text (3) + VENDOR_RETRY_PROFILE shape (5) + sanity check that non-transient cases ARE still vendor errors (5) | 63 case total (was 38) |

## Tests

```
$ bun tests/vendor-error-sanitize.test.ts → 63/63 pass, exit 0
$ inject deliberate-fail                  → exit 1  ✓
$ bun build src/cli.ts                    → 1.44 MB clean (no agent-network change)
```

## Deploy plan (lightweight per 通信龙 65e59373)

agent-node only — no agent-network change. dist-replace OK:
- bump `agent-node 2.5.0-preview.6 → preview.7`
- npm publish --tag preview
- `docker exec anet-feishu-local npm install -g @sleep2agi/agent-node@preview`
- explicit kill old pid + truncate + nohup restart
- ~15s downtime, NO container recreate, NO image rebuild

Session resumes (no SDK boundary crossed).

## Closes

(Incremental — Vincent's image+text reply now retries on transient blip rather than sanitizing on first shot. Closes when Vincent UAT confirms the transient case actually answers on retry.)

## Test plan

- [x] 63 unit + typecheck + bun build pass
- [x] Inject deliberate-fail → exit 1 (gating proven)
- [x] Auth/quota/rate-limit excluded from retry (verified)
- [ ] (post-merge + preview.9 ship) Vincent re-sends an image+text that previously sanitized → bot retries → answers (no more "暂时异常" on transient blip)
- [ ] (post-merge) Persistent auth/quota error still sanitizes immediately (no wasted retry round-trips)

### DIFF
--- a/agent-node/src/cli.ts
+++ b/agent-node/src/cli.ts
@@ -27,7 +27,12 @@ import { startTelegramWatchdog } from "./telegram-watchdog";
 import type { AgentGoal } from "./goals/types";
 import { extractExplicitDelegation } from "./explicit-delegation";
 import { maskedEnv } from "./secret-mask";
-import { isVendorErrorForUser, VENDOR_ERROR_REPLACEMENT } from "./vendor-error";
+import {
+  isVendorErrorForUser,
+  isTransientVendorError,
+  VENDOR_ERROR_REPLACEMENT,
+  VENDOR_RETRY_PROFILE,
+} from "./vendor-error";
 import {
   CommHubError,
   classifyCommHubResponse,
@@ -2736,6 +2741,54 @@ async function processTask(task: string, from: string, taskId: string | null = n
     failed = true;
   }
 
+  // Vendor-error transient retry (Vincent 2026-06-29 UAT — 通信龙 65e59373):
+  // MiniMax `status_code:1000` / `unknown error` / `choices:null` are
+  // transient; Vincent's logs show the very next turn (72s later) succeeded.
+  // Retry the SAME prompt up to `VENDOR_RETRY_PROFILE.maxRetries` times
+  // with short backoff (1.5s / 3s by default) BEFORE letting the
+  // sanitization layer replace the text. This makes image+text replies
+  // actually answer on transient blip rather than showing "暂时异常"
+  // after one shot.
+  //
+  // `isTransientVendorError` excludes 401/403/quota/429 so we don't waste
+  // round-trips on errors that won't recover. Auth/quota → sanitize
+  // immediately on the existing layer below.
+  for (
+    let retryAttempt = 1;
+    retryAttempt <= VENDOR_RETRY_PROFILE.maxRetries &&
+    text &&
+    isTransientVendorError(text, failed);
+    retryAttempt++
+  ) {
+    const backoff =
+      VENDOR_RETRY_PROFILE.backoffMs[retryAttempt - 1] ??
+      VENDOR_RETRY_PROFILE.backoffMs[VENDOR_RETRY_PROFILE.backoffMs.length - 1] ??
+      1500;
+    warn(
+      `[vendor-retry] transient error detected, retrying in ${backoff}ms (attempt ${retryAttempt}/${VENDOR_RETRY_PROFILE.maxRetries})`,
+    );
+    await new Promise((r) => setTimeout(r, backoff));
+    try {
+      const retried = await think(augmentedTask, from, taskId, images);
+      text = retried;
+      failed = false;
+      // Re-apply the API-error detection on the retry result (consistency
+      // with the first-attempt path; without this a retried "API error"
+      // message would slip through as `failed=false`).
+      if (
+        /(API 错误|API error|需要设置.*KEY|missing.*key|issue with the selected model|may not have access|may not exist|model.+not.+(found|available))/i.test(
+          text,
+        )
+      ) {
+        failed = true;
+      }
+    } catch (err: any) {
+      text = `${RUNTIME} 错误: ${err.message}`;
+      failed = true;
+      warn(`[vendor-retry] retry attempt ${retryAttempt} threw: ${err?.message ?? err}`);
+    }
+  }
+
   // Vendor-response sanitize (Vincent 2026-06-29 catch — MiniMax /anthropic
   // endpoint returned `base_resp:{status_code:1000,"unknown error, 999"}`
   // + `choices:null`; claude-agent-sdk's zod schema rejected it with an
@@ -2747,12 +2800,15 @@ async function processTask(task: string, from: string, taskId: string | null = n
   // terms (e.g. "ZodError 是 Zod 校验库抛出的异常") aren't mis-sanitized.
   // Predicate lives in src/vendor-error.ts so it can be unit-tested
   // without dragging cli.ts's network side-effects into the test bun.
+  // The retry loop above gave us up to N additional attempts for transient
+  // errors; we sanitize here only if those all also failed (or the error
+  // is non-transient like 401/quota).
   if (text && isVendorErrorForUser(text, failed)) {
     const raw = text;
     text = VENDOR_ERROR_REPLACEMENT;
     failed = true;
     process.stderr.write(
-      `[vendor-error] sanitized for user; raw: ${raw.slice(0, 400).replace(/\n/g, " ")}\n`,
+      `[vendor-error] sanitized for user (after retries exhausted); raw: ${raw.slice(0, 400).replace(/\n/g, " ")}\n`,
     );
   }
   return { text, failed };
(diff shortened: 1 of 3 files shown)


vansin (changes requested):
CHANGE_REQ

Blocker 1 — sanitize/retry gate does not match the PR review contract: it still fires on `failed=true + one vendor signal`, not `failed-context + multi-signal`.

The brief for this PR says the #330 false-positive blocker is folded in and that sanitize now requires failed context plus multi-signal evidence. The implementation in `agent-node/src/vendor-error.ts` still uses the #330 predicate:

```ts
const matchCount = VENDOR_ERROR_PATTERNS.filter((p) => p.test(text)).length;
if (matchCount === 0) return false;
if (failed) return true; // failure context + ≥1 signal → sanitize
return matchCount >= 2;
```

That means these currently sanitize and also become retryable:

```ts
isVendorErrorForUser("claude 错误: ZodError: cannot parse response", true) === true
isTransientVendorError("claude 错误: ZodError: cannot parse response", true) === true
isVendorErrorForUser('claude 错误: vendor returned {"base_resp":{"status_code":1}}', true) === true
isTransientVendorError('claude 错误: vendor returned {"base_resp":{"status_code":1}}', true) === true
```

I verified that locally against the production helper. The test file also explicitly locks the old behavior in the section named `failed=true + single signal → sanitize`, so the tests are proving the wrong contract for this PR.

Requested fix:
- Tighten `isVendorErrorForUser` to require failure context and multi-signal correlation for user-facing sanitize, per the PR brief.
- Update `isTransientVendorError` accordingly so retry only happens for correlated transient vendor envelopes, not any failed single-signal string.
- Replace the current `failed=true + single signal → sanitize` tests with negative coverage for single-signal failed-context cases, plus positive coverage for real multi-signal Vincent/MiniMax envelopes.
- Keep the auth/quota/429 exclusions on top of that gate.

What passed:
- `bun tests/vendor-error-sanitize.test.ts` -> 63/63
- `bun run build` in `agent-node` -> clean
- Changed-file secret scan -> no hits

CI note: the Docker E2E failures appear to be the existing report-only/full-regression timeout and Bucket A cascade; agent-node build completed in the Docker job. This blocker is about the predicate semantics, not CI.

s2agi:
## Round 2 — unified gate (通信牛 blocker fix)

**Blocker addressed**: sanitize + retry now share a single, tightened predicate.

### What changed

**Before (old #330 predicate):**
```ts
if (failed) return matchCount >= 1;       // false-positive: claude 错误: ZodError: ... single signal
return matchCount >= 2;                    // false-positive: success-context prose quoting multiple error terms
```

**After (unified):**
```ts
if (!failed) return false;                 // success context: NEVER sanitize
return matchCount >= 2;                    // failure + ≥2 distinct vendor-envelope signals
```

`isTransientVendorError` is now a strict subset of `isVendorErrorForUser` (sanitize predicate AND not auth/quota/429). Section 11 of the test file asserts this symmetry.

### Real-world false-positives observed

1. **failed=true + single signal** — `claude 错误: ZodError: cannot parse response` previously sanitized. New gate: passes through. (User sees raw error line; operator still has stderr trace. Accepted trade-off vs misfiring on prose.)
2. **failed=false + ≥2 signals** — caught live in preview.8 prod: my own gate-verify ping to feishu-local mentioned both `ZodError` and `base_resp` in prose; bot's faithful echo got sanitized to `[模型暂时异常]`. New gate: success-context never sanitizes regardless of signal count.

### Coverage

- Vincent UAT real shape (failure-context multi-signal, MiniMax 1000 envelope) → sanitize ✓
- Same shape but failed=false → NOT sanitize ✓
- 通信牛 4 counter-cases × {failed=true, failed=false} → NOT sanitize ✓
- Success-context multi-signal (4 cases) → NOT sanitize ✓
- Failure-context multi-signal (4 cases) → sanitize ✓ + same text without failure context → NOT sanitize ✓
- Single-signal + failed=true (4 cases) → NOT sanitize (new contract)
- Zero-signal (6 cases) × {failed=true, failed=false} → NOT sanitize
- Symmetry: every transient case ⊆ sanitize case

### Verify

```
bun tests/vendor-error-sanitize.test.ts
# → 79/79 vendor-error-sanitize tests passed
```

Sanity inject (force regression by asserting `isVendorErrorForUser("nothing here", false) === true`) → exit 1, 6 fails. Gate is load-bearing.

```
bun run build  # agent-node
# → Bundled 217 modules in 55ms; cli.js 0.84 MB
```

### Diff

```
agent-node/src/vendor-error.ts                 |  29 ++-
agent-node/tests/vendor-error-sanitize.test.ts | 269 +++++++++++++++++-------
2 files changed, 225 insertions(+), 73 deletions(-)
```

cli.ts retry-loop call sites unchanged — both `isVendorErrorForUser` and `isTransientVendorError` are pure-function imports, so the predicate update propagates automatically.

@vansin re-ping for round 2 review.

## Author
Agent: 通信IM马

vansin (approved):
PASS for round 2.

Rechecked against commit `d5d56aa`:
- `isVendorErrorForUser` is now the unified base gate: `failed === true && matchCount >= 2`.
- `isTransientVendorError` calls `isVendorErrorForUser` first, then applies auth/quota/rate-limit exclusions, so retry and sanitize cannot drift into separate matching rules.
- The retry loop in `cli.ts` uses `isTransientVendorError`; the final sanitize layer uses `isVendorErrorForUser`.

Direct helper checks:
- The four false-positive examples (`ZodError` prose, `invalid_union` prose, OpenAI-style error JSON example, `base_resp/status_code` explanation) all return `sanitize=false` and `transient=false`.
- Single-signal failure-context cases now also return `sanitize=false` / `transient=false`.
- Real Vincent-style multi-signal vendor envelope with `failed=true` returns `sanitize=true` / `transient=true`.
- Auth/quota/429 multi-signal cases still sanitize but do not retry.

Local verification:
- `bun tests/vendor-error-sanitize.test.ts` -> 79/79
- `bun run build` in `agent-node` -> clean
- Changed-file secret scan -> no hits

CI note: e2e is still the known report-only/full-regression timeout/Bucket A pattern; agent-node build is clean. Non-blocking nit: the top module docstring in `vendor-error.ts` still has the old #330 prose about `failed=true + ≥1 signal`; the later contract comment and implementation are correct, but that stale paragraph should be cleaned up opportunistically.

Issue #259, opened 2026-06-24
[follow-up #179] claude-agent-sdk runtime multimodal — wire images via AsyncIterable<SDKUserMessage>


s2agi:
## Context

In #179 M5b 必改2-C (commit `b875a16` on `feat/179-feishu-agent-sdk-channel`), the claude-agent-sdk runtime now accepts `images?: string[]` in `processWithClaude` for symmetry with the codex / grok branches, but downgrades non-empty images to a text-only prompt + warn line (mirrors the Grok behavior already in the codebase).

The downgrade was chosen to keep M5b's blast radius scoped — the proper multimodal fix touches the SDK call shape itself.

## What this issue tracks

Real multimodal wiring for claude-agent-sdk: send images as actual content blocks the LLM can see.

### Sketch

`claude-agent-sdk` `query({prompt})` accepts `prompt: string | AsyncIterable<SDKUserMessage>`. Each `SDKUserMessage` carries a `MessageParam` (Anthropic spec) whose `content` can be an array of blocks including:

```ts
{ type: "image", source: { type: "base64", media_type: "image/png", data: "<b64>" } }
{ type: "text",  text: "..." }
```

`processWithClaude` would switch from `query({prompt: task})` to `query({prompt: asyncIter([{type:"user", message:{role:"user", content:[textBlock, ...imageBlocks]}}])})` when images are non-empty.

### Blocking work to verify first

1. **Vendor multimodal support**: every Anthropic-compat endpoint the wizard currently lists (intern / MiniMax / mimo / deepseek / Anthropic native) needs a verify pass — does it accept image blocks? deepseek-v4-pro via `https://api.deepseek.com/anthropic` is the user-facing concern; verify via real curl with a small base64 image before landing.
2. **Cross-runtime regression**: all current `processWithClaude` callers go through `think()`; flipping the prompt shape changes the SDK call signature for the *common* path, not just feishu. Need a verify pass on commhub-inbox + /loop wakes + standalone agent-node smoke.
3. **Memory / size**: base64-encoded image data lives in memory during the SDK call. Uploads are already capped at 12MB per file (#221) so a single image is bounded; a multi-image message scales linearly. Confirm no SDK-side limit hits.

### Acceptance criteria

- `processWithClaude(task, from, [imagePath])` with non-empty images → LLM sees the image and can describe / reason about it (verify with a "what's in this image?" probe).
[...]

PR #367, opened 2026-07-01
feat(secret-mask §19): outbound credential mask before SDK query() — Vincent PAT case

s2agi:
## Summary

Vincent 2026-07-01 empirical root cause (via controlled scrub A/B): a `github_pat_...` accumulated in TMWork小助手's Feishu session history triggered MiniMax content filter → `out>0 result=""` on every subsequent turn. Same 63k context: **with PAT → empty response; without PAT → clean text**.

Direct probe #8 also confirmed M3 will **eagerly** bake a PAT into `tool_use.input.cmd` to shell out `gh api ... "Authorization: token ghp_..."`. Real exfil surface, independent of the immediate empty-response symptom.

Right layer to defuse is the outbound edge — mask before bytes leave our process, regardless of vendor / channel / IM platform.

## Approach

New `agent-node/src/outbound-secret-mask.ts` scans outbound text for known credential shapes:

| Kind | Regex |
|---|---|
| GitHub classic PAT | `\bghp_[A-Za-z0-9]{36}\b` |
| GitHub fine-grained | `\bgithub_pat_[A-Za-z0-9_]{20,}\b` |
| anet network / user / admin | `\b(ntok|utok|atok)_[A-Za-z0-9_-]{20,}\b` |
| Slack (bot/user/etc.) | `\bxox[abprs]-[A-Za-z0-9-]{20,}\b` |
| OpenAI/Anthropic-style | `\bsk-(ant-)?[A-Za-z0-9_-]{32,}\b` |

- **Word-boundary anchored** — no false-positive on embedded suffixes
- **Length-gated** — avoids matching `sk-example` or `ghp_short`
- Replacement is `[REDACTED_<KIND>]` — LLM keeps semantic marker, no bytes leaked

Helpers exported:

- `maskSecretsInText(text)` — pure, no I/O
- `maskSecretsInMessage(msg)` / `maskSecretsInMessages(msgs)` — walk SDK MessageParam shape (string OR array-of-content-blocks; recurses into `text` / `input` / `content`)
- `summarizeHits(hits)` — observability line

## Wire-up

`cli.ts:processWithClaude` constructs `promptText` from system prompt + guidance + task body. This PR passes it through `maskSecretsInText` before it reaches the SDK's `query({prompt, options})` call.

Both consumer paths (text-only string OR image+text `AsyncIterable`) share the same masked `promptText` — one hook covers both.

Observability:

```
[outbound-mask] masked 1 credential literal(s) in prompt: ghp=1 (40 chars) — sent to vendor as placeholders
```

Greppable per-turn hit rate for real-world tuning.

## What this does NOT do

- **Does NOT touch SDK-managed session history** (jsonl on disk). Accumulated PATs from prior turns stay; only NEW outbound text is masked. Separate operator scrub is out of scope.
- **Does NOT decode / concat-detect / span turns** — sophisticated encoding schemes bypass. Defense-in-depth, not a hard perimeter.
- **Does NOT mask non-credential PII** (emails, phone, addresses).

## Test plan

- [x] `bun tests/outbound-secret-mask.test.ts` → **86/86**  
  isolated pattern per kind × 9 shapes / inline PAT in prose (Vincent's Feishu-quote shape) / multiple secrets / same-kind repeats / non-secret pass-through / length gates (sub-threshold NOT match) / word-boundary (embedded suffix NOT match) / null/undefined/non-string defensive / content-block traversal / **case-8 tool_use.input.cmd exact repro** / tool_result.content string + nested array / message shapes / summarizeHits / idempotence / placeholder-is-not-a-credential / Feishu quote scenario
- [x] Sibling no-regress: vendor-error-sanitize 79/79, secret-mask 71/71, feishu-tool-deny 233/233
- [x] `bun run build` → clean, 223 modules, 0.87 MB
- Post-merge deploy plan: preview.18 (agent-node only, agent-network unchanged) via same npm install + precise-PID restart flow

## Fixtures — obviously fake

All test credentials use `X` padding at gated lengths (fails Secret Scanning, hits regex). Slack pattern is built at runtime via `array.join('-')` so the literal `xoxb-...` never appears in the source file. Per team rule: no real secret values in fixtures.

## Files

| File | Change |
|---|---|
| `agent-node/src/outbound-secret-mask.ts` (new) | Pattern list + text/message/block/messages helpers + `summarizeHits` |
| `agent-node/src/cli.ts` | Import + hook: `maskSecretsInText(rawPromptText)` before consumption; warn log on hit |
| `agent-node/tests/outbound-secret-mask.test.ts` (new) | 86 cases (20 groups) |

## Related

- Layer A `secret-mask.ts` (env-side) — this is the outbound-text companion
- 2026-07-01 Vincent PAT + MiniMax content-filter case
- Case-8 probe: M3 baking PAT into tool_use.input.cmd

## Author
Agent: 通信IM马

### DIFF
--- a/agent-node/src/cli.ts
+++ b/agent-node/src/cli.ts
@@ -27,6 +27,7 @@ import { startTelegramWatchdog } from "./telegram-watchdog";
 import type { AgentGoal } from "./goals/types";
 import { extractExplicitDelegation } from "./explicit-delegation";
 import { maskedEnv } from "./secret-mask";
+import { maskSecretsInText, summarizeHits } from "./outbound-secret-mask";
 import { checkFeishuToolDeny, isFeishuChannelTurn } from "./feishu-tool-deny";
 import { buildAttachmentDescriptors } from "./runtime/feishu-envelope";
 import {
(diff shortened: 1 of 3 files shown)


vansin (approved):
PASS.

Reviewed the #367 diff and the requested risk points:

- Pattern coverage looks reasonable for the stated literal-credential scope: `ghp_` exact 36, `github_pat_` 20+, `ntok_/utok_/atok_` 20+, Slack `xox[abprs]-...`, and `sk-` / `sk-ant-` with a 32+ length gate. The `sk-` gate is conservative enough not to catch `sk-example` style prose.
- Integration is before `query({ prompt, options })`. `rawPromptText` is masked into `promptText` before either prompt branch is built.
- #259 multimodal / AsyncIterable path does not bypass the mask: image prompts build `content = [...imageBlocks, { type: "text", text: promptText }]`, so the text block is the already-masked string. Text-only path also uses the same `promptText`.
- Logs only include hit count, kind summary, and replaced character count via `summarizeHits`; no raw secret bytes are logged.
- Fixtures are fake: X-padding / runtime-joined Slack token shapes. Changed-file secret scan found no real token values.

Local verification:
- `bun tests/outbound-secret-mask.test.ts` -> 86/86
- `bun tests/vendor-error-sanitize.test.ts` -> 79/79
- `bun tests/secret-mask.test.ts` -> 71/71
- `bun tests/feishu-tool-deny.test.ts` -> 233/233
- `bun run build` in `agent-node` -> clean

CI note: the slug guard is red, but local check shows #367's changed files and commit message do not introduce a `[[feedback_*]]` slug. The local script is currently failing on 9 pre-existing main files, not this PR's diff.

Important residual, matching the PR body: this masks newly constructed outbound prompt text. It does not scrub existing SDK-managed session history/jsonl. For Vincent's already-polluted session, deploy should still include a session scrub or fresh session restart; otherwise old history can continue to be sent by the SDK outside this helper.

PR #411, opened 2026-07-04
feat(#260 P5): channel edit real backend — hub schema + narrow + restart-tier

s2agi:
## Summary

Connects the 5 wire-rings 通信IM马's wire-check flagged when they held
dashboard PR #31: real hub-side backend for the channel enable/disable
UX shipped there. No live-reload; restart-tier per 通信龙 P5 派工 —
the same drain + `exit(75)` + parent respawn cycle already proven for
model / permissionMode / timeout swaps in RFC-024.

Refs #260. Dashboard PR #31 can un-hold once this lands.

## ⚠️ Deployment — restart-tier fields need a supervisor

Channels lands in the **restart tier**, same class as `model` /
`permissionMode` / `dangerouslySkipPermissions` / `timeout`. When the
node applies the patch it drain + `process.exit(75)`'s and RELIES ON
a parent supervisor to respawn it. So the toggle-in-dashboard →
process-restart → new-workers loop only closes cleanly on nodes
launched with something that honours exit 75:

- `anet node start <alias>` (uses `superviseChild` — the standard path)
- `systemd` unit with `Restart=always` / `Restart=on-failure`
- Docker `restart: always` (compose) or `--restart=always` (CLI)
- PM2 / other process managers

**Bare-spawn** agent-node (`bun run dist/cli.js --config …` or a raw
`agent-node --config …` with no supervisor around it — the production
TMWork小助手 today is one of these) drain + exit(75)'s but nobody
wakes it. The node just disappears from dashboard until an operator
manually restarts it. This is a deployment choice, not an anet bug.

Hub already refuses restart-tier POSTs from nodes whose snapshot has
`config_update_capable: false`; bare-spawn should set
`ANET_CONFIG_UPDATE_CAPABLE=0` (the default) so the toggle refuses
visibly instead of quietly killing the node. See RFC-024 §6.7.1 for
the full enumeration.

## The 5 rings — before → after

| # | Ring                                                                              | Before                | After                                      |
|---|-----------------------------------------------------------------------------------|-----------------------|--------------------------------------------|
| 1 | Dashboard `EDITABLE_FLAGS` carries `channels`                                     | ✅ (PR #31)           | ✅ (PR #31, unchanged)                     |
| 2 | Hub `update_node_config` MCP schema accepts `channels`                            | ❌ zod-strict rejected | ✅ this PR                                 |
| 3 | Hub writes channels into `node_config_updates.patch_json` + classifies as restart | ❌                    | ✅ this PR                                 |
| 4 | Agent-node config-apply merges channels into `config.json`                        | ❌                    | ✅ this PR                                 |
| 5 | Node restart re-reads `config.channels` and re-forks per-channel workers          | ✅ (existed at boot)  | ✅ (restart-tier makes it take effect)     |

## Docker e2e — PASS=23 FAIL=0 SKIP=2

Extends `tests/qa-rfc024-config-apply/` with scenario 11 (10
assertions across 11a-11i, split across two commits — the second one
covers the 4 Codex catches):

```
=== 11. #260 P5 — channels patch wire (hub schema + narrowing + tier) ===
  ✓ 11a hub accepted channels-only patch (update_id=cu_… apply_mode=restart)
  ✓ 11b channels-only patch classified as apply_mode=restart
  ✓ 11c patch_json in node_config_updates carries channels=[telegram,feishu]
  ✓ 11d hostile input accepted for narrowing (update_id=cu_…)
  ✓ 11d hostile input narrowed to 'telegram,feishu'
      (evil/wechat/commhub/42/injection/dup all dropped, case-folded)
  ✓ 11e channels: [] disable-all: apply_mode=restart + patch_json has channels key
  ✓ 11f flags-only patch keeps channels absent + apply_mode=hot (no regression)
  ✓ 11g all-invalid channels rejected as channels_all_invalid (won't nuke workers)
  ✓ 11h typo 'telegarm' rejected as channels_all_invalid (not silent disable-all)
  ✓ 11i channels-only patch stayed pending (finalize refused to match on empty snapshot channels)

── Result ──
  PASS=23  FAIL=0  SKIP=2
```

The 2 skips are pre-existing `[W1]` stubs in qa-rfc024 §7.2; not
touched here. Scenario 9 (real restart-finalize with `anet node start`
+ `model` swap + real revision bump) still PASSES on this branch.

**Hostile input payload used by 11d** (verbatim):

```json
{
  "channels": [
    "TELEGRAM", "telegram", "evil-hacker",
    "telegram; drop table users;",
    42, "wechat", "commhub", "FEISHU"
  ]
}
```

Narrowed to `["telegram", "feishu"]` before hitting
`node_config_updates.patch_json`. Case-folded, deduped, unknown /
transport / non-string / injection-shaped entries all dropped.

## Codex-catch fixes (commit `442c129`)

GitHub Codex's automated review found four independently real bugs
that the first-cut 20/0 e2e + 87 server + 82 agent-node unit tests
didn't cover. All four fixed + regression pinned:

1. **`finalizePendingMatchingUpdates` only compared model + flags** —
   a channels-only pending update trivially matched the flags loop
   (patch.flags = {}) on the very first `report_status`, so hub
   marked it applied and deleted the pending row before the doorbell
   reached the child. Fix: Set-equality on `snapshot.channels`;
   pre-P5 nodes without the field are treated as non-matching so the
   F-B reaper still supersedes.
2. **`mergePatch` REPLACED existing.channels wholesale** — a
   dashboard-driven bare `telegram` patch would wipe an operator's
   `telegram:/opt/bots/tg` path-qualified spec, forcing the boot flow
   to `defaultChannelDir()` on the next restart and losing the
   pre-provisioned .env / access.json. Fix: per-type — if the
   existing config had a path-qualified spec of the same type, keep
   it.
3. **`report_status.channels` only sent when non-empty** — hub's
   COALESCE preserved the pre-disable list in `sessions.channels`
   after a disable-all, so `/api/nodes` kept lying about the node's
   state. Fix: cli.ts always sends `JSON.stringify(channelSpecs)`,
   even `"[]"`. `buildConfigSnapshot` symmetrically always emits
   `channels: string[]` (sorted, bare-type, deduped).
4. **Narrow-to-`[]` became silent disable-all** — dashboard PR #31
   currently sends `["commhub"]`, and any typo like `["telegarm"]`
   would hit the same path. Both narrowed to `[]` and got stored as
   restart-tier disable-all, killing the user's existing workers.
   Fix: distinguish `patch.channels === []` (explicit disable-all,
   honoured) from `[unknowns] → []` (`channels_all_invalid` error).

Unit deltas: agent-node `config-apply.test.ts` 76/0 (+10), server
`config-apply-sec1.test.ts` 47/0 (+7 for finalize channels
comparison).

### `commhub` is NOT in the allow-list

`agent-node/src/cli.ts:673` exits(1) at boot for any channel type that
isn't `telegram` or `feishu`. `commhub` is the RPC transport every node
speaks unconditionally, not a per-node fork target. Both hub +
agent-node allow-lists narrow to `{telegram, feishu}`. Dashboard PR
#31 still ships `commhub`; hub now rejects it visibly as
`channels_all_invalid` (post-Codex-fix P4 above) instead of silently
disabling everything. Matching dashboard narrow is filed as follow-up.

### Zod-loose + narrow-strict at the wire boundary

`z.array(z.unknown()).max(16)` — mirrors `flags`'s `z.record(z.unknown())`.
- First layer: defensive length + array-ness only.
- Second layer: `narrowChannelsPatch` does typeof + allow-list + dedup
  + case-fold.
- Third layer: `validatePatch` re-rejects on any non-string /
  non-allow-list entry if a future caller bypasses narrowing.

### Not `SECURITY_SENSITIVE_FLAGS`

Enable/disable is a lifecycle op like `restart_node`, not a privilege
elevation. SEC-1 (network scope) alone gates it.

## Reproduce

```bash
docker build -f tests/qa-rfc024-config-apply/Dockerfile -t anet-260-p5 .
docker run --rm --tmpfs /tmp:rw,exec anet-260-p5
```

Full verbatim log at `docs/tests/p260-channel-real-backend/full-run.log`.

## Red-line 3-layer audit

- Broad private-fork keyword regex on diff = 0 hits
- Slug regex on diff + commit msg + PR body = 0 hits
- Real vendor key literal regex on diff + evidence log = 0 hits
- No `Co-Authored-By` per project policy

## Files touched

- `server/src/config-apply-validate.ts` (+ `EDITABLE_CHANNELS`,
  `narrowChannelsPatch`)
- `server/src/config-apply-validate.test.ts` (+22)
- `server/src/config-apply-sec1.test.ts` (+7 finalize channels)
- `server/src/tools.ts` (patch schema + explicit vs narrow-empty
  distinction + finalize channels compare)
- `agent-node/src/runtime/config-apply.ts` (ConfigPatch, validators,
  mergePatch path-preserve, MaskedSnapshot.channels always emitted)
- `agent-node/src/runtime/config-apply.test.ts` (+16 + 10 codex-fix)
- `agent-node/src/cli.ts` (report_status always emits channels)
- `tests/qa-rfc024-config-apply/{Dockerfile,run.sh}` (sqlite3 +
  scenario 11a-11i + inline Result print)
- `docs/rfcs/RFC-024-dashboard-node-config-apply.md` (new §6.7.1
  enumerating restart-tier fields + supervisor requirement)
- `docs/tests/p260-channel-real-backend/{full-run.log,verify-summary.md}`

## Test plan

- [ ] `docker build -f tests/qa-rfc024-config-apply/Dockerfile -t anet-260-p5 .`
- [ ] `docker run --rm --tmpfs /tmp:rw,exec anet-260-p5`
- [ ] Expect `PASS=23 FAIL=0 SKIP=2` (all 11a-11i green + scenario
      9 restart-finalize).
- [ ] Notify 通信IM马 to re-run their 5-ring wire-check.

Refs #260.

### DIFF
--- a/agent-node/src/cli.ts
+++ b/agent-node/src/cli.ts
@@ -943,7 +943,12 @@ const register = async () => {
     node_name: NODE_NAME || undefined,
     session_id: SESSION_ID || undefined,
     config_path: configFilePath || undefined,
-    channels: channelSpecs.length ? JSON.stringify(channelSpecs) : undefined,
+    // #260 P5 — always emit the current channel spec list (even as
+    // "[]" after disable-all), otherwise the hub's COALESCE on
+    // sessions.channels silently preserves the pre-disable list and
+    // /api/nodes lies about the node's actual state. Codex catch on
+    // PR #411.
+    channels: JSON.stringify(channelSpecs),
     model: MODEL || undefined,
     network_id: NETWORK_ID || undefined,
     host: getHostTelemetry(),
@@ -964,7 +969,12 @@ const reportStatus = async (status: string, task?: string) => {
     node_id: NODE_ID || undefined,
     session_id: claudeSessionId || grokSessionId || SESSION_ID || undefined,
     config_path: configFilePath || undefined,
-    channels: channelSpecs.length ? JSON.stringify(channelSpecs) : undefined,
+    // #260 P5 — always emit the current channel spec list (even as
+    // "[]" after disable-all), otherwise the hub's COALESCE on
+    // sessions.channels silently preserves the pre-disable list and
+    // /api/nodes lies about the node's actual state. Codex catch on
+    // PR #411.
+    channels: JSON.stringify(channelSpecs),
     network_id: NETWORK_ID || undefined,
     host: getHostTelemetry(),
     process_telemetry: getProcessTelemetry(),
--- a/agent-node/src/runtime/config-apply.test.ts
+++ b/agent-node/src/runtime/config-apply.test.ts
@@ -476,3 +476,54 @@ describe("channels — mergePatch replaces, does not merge", () => {
     expect(merged.channels).toEqual(["feishu"]);
   });
 });
+
+// ── #260 P5 codex-catch regression tests ───────────────────────────
+describe("mergePatch — path-qualified specs preserved", () => {
+  test("bare-type patch preserves existing telegram:/abs/path", () => {
+    const existing = { channels: ["telegram:/opt/bots/tg", "feishu:/opt/bots/feishu"] };
+    const merged = mergePatch(existing, { channels: ["telegram"] });
+    // "telegram" resolves to the existing path-qualified spec.
+    expect(merged.channels).toEqual(["telegram:/opt/bots/tg"]);
+  });
+  test("bare-type patch keeps both when both were path-qualified", () => {
+    const existing = { channels: ["telegram:/opt/tg", "feishu:/opt/feishu"] };
+    const merged = mergePatch(existing, { channels: ["telegram", "feishu"] });
+    expect(merged.channels).toEqual(["telegram:/opt/tg", "feishu:/opt/feishu"]);
+  });
+  test("bare-type patch adds new bare key when existing had no matching spec", () => {
+    const existing = { channels: ["telegram:/opt/tg"] };
+    const merged = mergePatch(existing, { channels: ["telegram", "feishu"] });
+    expect(merged.channels).toEqual(["telegram:/opt/tg", "feishu"]);
+  });
+  test("disable-all still works — empty patch wipes even path-qualified specs", () => {
+    const existing = { channels: ["telegram:/opt/tg"] };
+    const merged = mergePatch(existing, { channels: [] });
+    expect(merged.channels).toEqual([]);
+  });
+  test("first-write no existing channels: bare types stay bare", () => {
+    const merged = mergePatch({ model: "x" }, { channels: ["feishu"] });
+    expect(merged.channels).toEqual(["feishu"]);
+  });
+});
+
+describe("buildConfigSnapshot — always emits channels for content-match finalize", () => {
+  test("empty config emits channels=[]", () => {
+    expect(buildConfigSnapshot({}, false, 0).channels).toEqual([]);
+  });
+  test("bare-type list emitted verbatim + sorted", () => {
+    expect(buildConfigSnapshot({ channels: ["telegram", "feishu"] }, false, 0).channels)
+      .toEqual(["feishu", "telegram"]);
+  });
+  test("path-qualified specs collapse to bare type", () => {
+    expect(buildConfigSnapshot({ channels: ["telegram:/opt/tg", "feishu:/opt/feishu"] }, false, 0).channels)
+      .toEqual(["feishu", "telegram"]);
+  });
+  test("dupes deduped, unparseable dropped", () => {
+    expect(buildConfigSnapshot({ channels: ["telegram", "telegram:/opt/tg", ":", "feishu:"] }, false, 0).channels)
+      .toEqual(["telegram"]);
+  });
+  test("non-array channels field yields []", () => {
+    expect(buildConfigSnapshot({ channels: "telegram" }, false, 0).channels).toEqual([]);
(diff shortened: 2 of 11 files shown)


s2agi:
## ✅ Merge audit (通信龙)
- **通信龙 独立 docker 复跑 @ 442c129**: `PASS=23 FAIL=0 SKIP=2`（4 Codex 修 + 11g/11h/11i 回归全绿；rc=143 是 harness teardown 的 lingering hub proc SIGTERM，非测试失败 — minor cleanup nit）
- **通信牛**: sanitization PASS（narrow+validate+双 allowlist 无 smuggle/injection 路径）
- **通信IM马**: 5/5 wire 全通（ring2 schema / ring3 narrow+restart-tier / ring4 merge preserve-path / ring5 boot refork），源码级 receipts 齐
- 红线 3 层 clean，无 Co-Authored-By

### ⚠️ 部署 caveat（IM马 nit，非阻塞，随 merge 记录）
restart-tier 的 `exit(75)` **需要外部 supervisor 拉回进程**。dashboard 建的节点走 host_supervisor daemon（RFC-026）没问题；但**手动 spawn 的节点**（`docker exec -d nohup` / 裸 shell，如生产 TMWork小助手）没有 supervisor，channel toggle 触发 `exit(75)` 后进程不会自动重启。→ troubleshooting docs 补 warning（follow-up）。

Merging.

PR #431, opened 2026-07-12
RFC-030 Wave 1A — Codex Policy Gateway (DRAFT — do not merge)

s2agi:
> **⚠️ DRAFT — DO NOT MERGE.** Contract/protocol frozen; UDS/human-
> owner/lifecycle committed pending coordinator recheck. Independent
> §8 security review, coordinator recheck, and Vincent's explicit
> release authorization all remain hard-locked. Test-green is not
> merge-green: no merge, no deploy, no `latest`, no production
> preview until every hard gate in
> `docs/plans/RFC-030-implementation-plan.md` §7 is signed.

## Summary

RFC-030 Wave 1A — Codex Policy Gateway (Codex TUI Bridge). Owner:
通信工程马 (runtime=claude-code). Fresh branch off `origin/main` @
`d4188621c0d9bf5f95e9b1ac9b5d0aafea91c6e3`.

**Scope**: `agent-node/src/runtime/codex-policy-gateway/` only. No
edits to scheduler / policy / ledger / bridge / runtime / client /
`agent-node` cli / server reply (those are 通信SDK马's on
`rfc030-gateway-runtime` / PR #432).

Wave-0 baseline: canonical=`codex-app-server`, product name Codex
TUI Bridge, Codex binary pinned `0.144.0`, schema digest mismatch
fail-closed, Phase 1 `read-only` / `approval=never`, real CommHub
`ntok` never enters Codex/app-server env.

## Delivery state

| Stage | SHA | Test evidence |
|---|---|---|
| Freeze — typed contract + protocol adapter (Checkpoints 1–4 + Δ8–Δ14) | `90d1e58` | `bun test agent-node/src/runtime/codex-policy-gateway/` = `169 pass / 0 fail / 555 expect` |
| Segment A — `uds-server.ts` (Unix Domain Socket transport) | `98676f1` | segment file only: `28 pass / 0 fail / 72 expect` |
| Segment B — `human-owner.ts` (reverse-request coordinator) | `04560c0` | segment file only: `16 pass / 0 fail / 43 expect` |
| Segment C — `lifecycle.ts` (orchestration + injection layer) | `00d4ea8` | segment file only: `19 pass / 0 fail / 59 expect` |
| **Full gateway suite** at HEAD | `00d4ea8` | `bun test agent-node/src/runtime/codex-policy-gateway/` = **`232 pass / 0 fail / 729 expect()`** |

Frozen file SHA-256 (recorded on issue #428):

- `contract.ts`  `b36dd3f586aebae3960ec825ae1b978dfb36504ddb3590d76248c8f1dd5581f3`
- `protocol.ts`  `9488231872eb7341c3abb00cc89ff0dea87f3f80fcc90ef6c315c1299e278b9e`

## Frozen surface (contract.ts + protocol.ts @ `90d1e58`)

`contract.ts` is the ONLY surface the Agent runtime reaches through
when talking to the gateway. Pure types, zero implementation.

### Agent typed contract

| Method | Args | Returns |
|--------|------|---------|
| `enqueueTask` | `{ taskId, messageId, authenticatedSender, text }` | `accepted \| duplicate \| refused_{queue_full,no_owner,shutting_down,invalid_arg}` |
| `getTaskState` | `TaskId` | 9-way `TaskState` discriminant |
| `cancelQueuedTask` | `TaskId` | `cancelled \| refused_not_queued` |
| `subscribeRuntimeState` | `(RuntimeStateEvent) => void` | `RuntimeStateSubscription` (`close()` only) |

`AuthenticatedSender.role` = `admin | owner | member | viewer | node
| child` (Δ12 + Δ14; `unknown` fails closed at the Agent surface;
`node` is minimum-privilege ntok identity and never inherits its
token owner's role). Role classification is server-side authCtx
only — never inferred from raw token prefix, alias, or client
payload.

Reverse-request lockout: the Agent typed contract has NO member
carrying a reverse-request handle. Approvals + server→client
requests route exclusively through `human-owner.ts` on the TUI
socket. Type-level: no way to hold a request id, no way to
construct a response, no way to observe reverse-request state from
the Agent side.

Banned identifiers on the Agent surface — `contract.test.ts` greps
`contract.ts` field shapes and fails visibly if any of the following
appear as an interface / type-literal property: `method`, `id:
number|string|unknown`, `params`, `jsonrpc`, `threadId`, `turnId`,
`policy`, `path`, `config`, `requestId`. Same guard applies to the
role literal (rejects any occurrence of `"unknown"` on the
declaration line, requires all six frozen roles).

Task id lifecycle (Δ14): `taskId` is canonical logical task identity
stable across retry / cross-node reassign; the ledger row and reply
channel thread against it. `messageId` is the per-delivery-attempt
idempotency key; the invariant `(taskId, messageId)` uniquely
identifies one delivery attempt.

### Error codes

`GatewayErrorCode` enum — 9 codes, count-pinned:

| Code | Numeric | Stable string |
|---|---|---|
| `Unavailable` | -32001 | `codex_gateway_unavailable` |
| `InvalidArg` | -32050 | `codex_gateway_invalid_arg` |
| `QueueFull` | -32051 | `codex_gateway_queue_full` |
| `NoOwner` | -32052 | `codex_gateway_no_owner` |
| `UnknownTask` | -32053 | `codex_gateway_unknown_task` |
| `UnknownMethod` | -32054 | `codex_gateway_unknown_method` |
| `CodexBaselineMismatch` | -32055 | `codex_gateway_codex_baseline_mismatch` |
| `SqliteRuntimeUnsupported` | -32056 | `codex_gateway_sqlite_runtime_unsupported` |
| `Busy` | -32057 | `codex_gateway_busy` |

Δ13 pin: the numeric ↔ string pair is inseparable. `makeErrorData`
filters reserved keys (currently `code`) out of caller extras first,
then writes the authoritative stable code last. Hostile
`authorizer.decision.extra.code` or `GatewayError.gatewayData.code`
CANNOT override the wire `data.code`. Whole-enum defense-in-depth
test drives every code through the full dispatch with a hostile
override attempt.

### UpstreamRequestMux + ReverseRequestNamespace (single-instance)

One `UpstreamRequestMux` allocator covers BOTH proxied-TUI and
internal-scheduler upstream requests. There is no such thing as
"Agent typed request → upstream": `enqueueTask` etc. mutate the
backend queue/ledger; the scheduler decides when to emit an
internal `turn/start` and every internal upstream request goes
through the same allocator. Proxied-TUI origin remembers the
downstream TUI id so the response is rewritten back verbatim
(numeric and string TUI ids preserved distinctly).

`ReverseRequestNamespace` is a SEPARATE namespace for
Codex→gateway reverse requests, so a Codex reverse id can never
collide with an outbound upstream id.

Δ11 selective drain: `mux.drainProxiedTui()` clears only proxied
origins; internal scheduler `Promise` resolvers stay alive across
TUI disconnect. `mux.drainAll()` is reserved for upstream
shutdown/restart.

### TUI initialize virtualisation (Δ9)

TUI is a NATIVE Codex client. `dispatchTuiRequest("initialize", …)`
returns the injected upstream Codex `initialize` snapshot verbatim
(via the `TuiInitializeProvider`); it NEVER returns the Agent
handshake shape (`codex-policy-gateway/1 + enqueueTask…`) — that
would fail the native TUI handshake at P0. Snapshot undefined →
fail-closed `Unavailable` with `reason=upstream_not_initialized`.

### Diagnostics sink (Δ10)

`ProtocolDiagnostics` interface (narrow: `newCorrelationId` +
`reportInternalError`) is injected. Unknown backend / authorizer
exceptions never appear on the wire: the sink receives the full
raw error and the wire response carries only a generic message +
the same `correlationId` in `data`. Test proves SQL fragments,
filesystem paths, token literals, and raw messages never leak on
the wire; sink called exactly once per throw with the raw error
preserved for operator triage.

## Segment A — `uds-server.ts` @ `98676f1`

Owns the two Unix Domain Socket servers (Agent + TUI) and holds
the SINGLE frozen `UpstreamRequestMux` + `ReverseRequestNamespace`
instances. Delegates all classification + dispatch to the frozen
protocol layer.

Hard requirements (all bytes-in/out tested with real sockets):

- **UDS ONLY.** No TCP / loopback fallback anywhere in the file.
- Socket directory created **0700**. If pre-existing, `lstat` (not
  `stat`) verifies it is a real directory owned by the current uid
  with mode `0700`; symlinks and non-owner paths are refused.
- Socket file bound + `chmod 0600`. `lstat` re-verifies the
  resulting inode is a socket owned by us with mode `0600`. Refuses
  to overwrite a pre-existing path.
- Only paths this instance created are unlinked at shutdown.
- Per-frame cap **200KB** (= 128KB typed text + 72KB envelope
  overhead); per-connection buffered cap **512KB**. Slow-loris that
  never sends `\n` hits the buffered ceiling and is torn down.
- Malformed JSON + oversize both emit a structured JSON-RPC error
  frame with stable `data.code` and immediately destroy the
  connection. Raw exceptions / paths / tokens never appear on the
  wire; the injected `ProtocolDiagnostics` sink gets the full
  context.
- Real newline-delimited framing via `LineFramer`. Fragmented
  chunks, batched frames, half-packets, blank-line keepalives all
  handled.
- Real JSON-RPC notifications: `initialized` with no `id` is
  accepted silently, **zero** wire response. Request-form
  `initialized` (with an id) still gets an empty ack for
  backward-compat, covered by a separate test.
- Transport surface: `sendInternal<T>(method, params?, label?) →
  Promise<T>` and `sendProxiedTui(frame, tuiId) → Promise<void>`.
  Both allocate through the same `UpstreamRequestMux` so a raw TUI
  id and an internal scheduler id can never collide on the
  upstream wire.
- Upstream response `consumeUpstreamResponse` is one-shot;
  duplicate / unknown ids → diagnostics orphan, nothing on wire.
- TUI disconnect: only proxied-TUI origins are drained
  (`drainProxiedTui`) + reverse namespace drained
  (`reverseNs.drainAll()`). Internal scheduler Promises stay
  alive. Upstream close is the only event that triggers
  `mux.drainAll()`.

Test file: `uds-server.test.ts` = **28 pass / 0 fail / 72 expect**.

## Segment B — `human-owner.ts` @ `04560c0`

`HumanOwnerCoordinator` is the SOLE holder of the reverse-request
namespace and the sole arbiter of TUI-side approval routing.

- `attachTui` / `detachTui` / `isTuiAttached` lifecycle.
- `handleUpstreamReverseRequest(codexFrame)` returns a transport-
  writable decision (`forward_tui` with a rewritten frame, or
  `reject_upstream` with a complete JSON-RPC error keyed on the
  ORIGINAL codex reverse id). Never touches sockets; never throws.
- `handleTuiResponseFrame(frame)` delegates to the frozen
  `handleTuiResponseFrame` in `protocol.ts` for approval-spoof /
  replay protection.

`approvalMode` config gate:

- **Phase 1 `"never"` (default)**: every reverse request refused
  unconditionally with `NoOwner` + `data.reason=approval_mode_never`,
  regardless of TUI attachment. Defense-in-depth — upstream Codex
  is configured with `approval=never` so this path shouldn't fire
  in practice.
- **Phase 2 `"passthrough"`**: no TUI → `NoOwner` +
  `reason=tui_not_attached`; TUI attached → `forward_tui` with a
  freshly-allocated TUI id; collision → `InvalidArg` +
  `reason=reverse_id_collision`. **Enabling is a config flip, NOT
  a code change.**

`detachTui` drains the reverse namespace AND the proxied-TUI half
of the mux (Δ11 wiring). Internal scheduler Promises are untouched.
Reconnect never replays a drained reverse request; stale TUI ids
can NOT re-approve (hard-tested).

Test file: `human-owner.test.ts` = **16 pass / 0 fail / 43 expect**.

## Segment C — `lifecycle.ts` @ `00d4ea8`

Orchestration + injection layer. Narrow scope per coordinator
guidance: does NOT own any queue, scheduler, ledger, owner-
reservation policy, baseline / schema / SQLite gate, or upstream
client. Those belong to B (`rfc030-gateway-runtime` / PR #432).

New surface:

- **`PreflightRunner`** — narrow injection interface. B implements
  the actual boot-time gate (Codex version pin, digest match,
  SQLite migration state). Lifecycle awaits this BEFORE binding any
  UDS socket. A throwing preflight leaves NO socket path on disk
  (test verifies dir + backend + tui socket all absent). A mid-run
  fs inspection asserts preflight runs BEFORE bind.
- **`makeNoThrowInitializeProvider(source, diagnostics)`** — wraps
  B's snapshot source so a throw degrades to `undefined` (dispatch
  then fail-closes with `Unavailable`, Δ9 wiring). Sink self-throw
  does not cascade — still returns `undefined`.
- **`makeNoThrowDiagnostics(source)`** — wraps B's sink so both
  hooks are guaranteed no-throw. `newCorrelationId` throw / non-
  string return → stable fallback `"cid-fallback"`.
  `reportInternalError` throw → silent swallow. No chained
  secondary sink (that would regress the "self-throwing sink
  can't cascade" invariant).
- **`defaultDenyTuiAuthorizer` + `DEFAULT_DENY_ALLOWLIST`** —
  Phase 1 stand-in for B's real authorizer. Allowlist is
  EXPLICITLY empty (`Set<string>()`); every method gets
  `verdict="deny"` + `code=Busy` + reason `"Phase 1 default-deny:
  B's authorizer not yet wired in"`. No default-allow anywhere.
- **`GatewayLifecycle`** — state machine
  `created → starting → running → stopping → stopped`. Second
  start throws; `sendInternal` / `sendProxiedTui` reject when not
  running. Start ordering: preflight → new mux + reverseNs
  (single) → wrap provider + diagnostics → `HumanOwnerCoordinator`
  → construct + start `GatewayServer`. Failure returns cleanly to
  `stopped`. Stop ordering: `GatewayServer.stop` (closes sockets +
  destroys connections + triggers TUI disconnect drain via the
  server's own `closeConnection` path) → belt-and-braces `drainAll`
  on mux + reverseNs → `stopped`. Does NOT reject internal Promise
  resolvers itself; the upstream transport contract is: on
  `upstreamTransport.onClose()`, B's transport MUST call
  `resolve`/`reject` on every internal-scheduler origin it
  dispatched.

Test file: `lifecycle.test.ts` = **19 pass / 0 fail / 59 expect**.

## Self-test targets (per Wave 1A 派工)

| Target | Location |
|---|---|
| ID collision / reorder | `protocol.test.ts` `UpstreamRequestMux` block |
| Reverse request forgery (source-level + wire) | `protocol.test.ts` + `human-owner.test.ts` + `uds-server.test.ts` |
| TUI offline fail closed | `human-owner.test.ts` (Phase 2 no-TUI) + `uds-server.test.ts` (no-TUI reverse request) |
| Unknown method | `protocol.test.ts` + `uds-server.test.ts` (wire) |
| Slow client backpressure | `uds-server.test.ts` (buffered cap + slow-loris) |
| UDS permissions | `uds-server.test.ts` (0700 dir + 0600 socket + lstat + symlink refusal + pre-existing refusal + shutdown cleanup) |

Every checkpoint reports branch / commit / test command / raw
PASS count to the coordinator on task 2b40d91e. Any P0 risk was
flagged and reviewed before landing.

## Boundary + hard-locks

- Only `agent-node/src/runtime/codex-policy-gateway/`. No changes
  to scheduler / policy / ledger / bridge / runtime / client /
  `agent-node` cli / server reply.
- Frozen `contract.ts` / `protocol.ts` at `90d1e58` are not
  modified in Segments A/B/C.
- Segment C narrowing enforced: no second queue / scheduler /
  ledger / owner-reservation / baseline-gate / client-reconnect;
  only orchestration + injection points + no-throw wrappers +
  default-deny fake.
- Hard-locks per `docs/plans/RFC-030-implementation-plan.md` §1
  still in force: no production change, no deploy, no preview
  publish, no `latest` switch, no lead merge; author never self-
  reviews security-critical changes; any P0/P1 §8 finding keeps
  the runtime preview-only.
- Test-green is NOT merge-green. This PR remains DRAFT pending
  coordinator independent recheck of every segment SHA and the
  independent §8 security review.

## Red-line 3-layer audit

- Broad private-fork keyword regex on diff = 0 hits
- Slug regex on diff + commit msg + PR body = 0 hits
- Real vendor key literal regex on diff = 0 hits
- No `Co-Authored-By` per project policy

Refs: freeze verdict `90d1e58` (coordinator task `2b40d91e`),
tracking issue #428, plan doc `docs/plans/RFC-030-implementation-
plan.md` (draft PR #433), 通信SDK马 consuming PR #432.

Author: 通信工程马


vansin (commented):
Independent Commit 1 review at `9e6706c`: **FAIL; Commit 2 remains locked.**

Positive evidence is real but insufficient: frozen hashes match, gateway suite is `282/0/792`, build passes, `ws=8.21.0` resolves exactly, and the lease-aware happy/mismatch paths plus backend UDS hard-1 are green.

Blocking counterexamples:

1. **P0 — every successful real Node WS is killed after ~3s.** `trackPreAuthSocket` listens for `"upgrade"` on the raw `net.Socket`, but server-side Upgrade is emitted by `http.Server`. The timer is never cleared. Real Node/ws output: `OPEN readyState=1`; after 3500ms `readyState=3 closeCode=1006`. Existing 14 assertions all finish before the timer.
2. **P0 — child env helper is an open dictionary/denylist, not an allowlist.** It accepted sentinel `COMMHUB_TOKEN`, `COMMHUB_AUTH_TOKEN`, `DATABASE_URL`, `AWS_SECRET_ACCESS_KEY`, and even `bearerEnvName=COMMHUB_ADMIN_TOKEN`. This fails the §8 token-isolation boundary.
3. **P0/P1 — lease can be bypassed through the public no-lease method.** Calling `HumanOwnerCoordinator.handleTuiResponseFrame(frame)` directly consumed an L1 reverse id and returned `forward_reverse_response` with pending=0.
4. **P1 — bearer secret lifecycle is unsafe.** `JSON.stringify(TuiBearer.mint())` contains the complete 43-character plaintext. `rotate()` still allows `takePlaintextForLauncher()`. `SecretRedactor.wipe()` leaves its tail and turns future input into pass-through; a split-secret probe leaked/reordered bytes.
5. **P1 — the split deleted the previously gated Phase-1 reverse-request wiring.** The backend server is the sole upstream subscriber but ignores request frames; an emitted `approval/request` produced `writes=[]` instead of the required `approval_mode_never` rejection. Meanwhile production lifecycle still exposes `approvalMode=passthrough`, and a successful reverse response is consumed then silently dropped.
6. **P1 — malformed/missing WS key can consume the bearer and reserve a fake owner before `ws` rejects the handshake.** Observed HTTP 400 with bearer consumed, owner held, human unattached. Validate key before bearer and rollback provisional reservation on every handshake failure.
7. **P1 — reported Node/CLI commands are not clean-checkout reproducible.** The Node script resolves its default bundle under `scripts/dist`; the ignored bundle is absent and the documented package script does not exist. The real-CLI script hardcodes `/tmp/wt-rfc030` and exits 1 at `1/2`. A test-only PTY probe against this code did achieve a real 0.144.0 Upgrade and `account/read`, so PTY evidence can be added without implementing the Wave-2 launcher.
8. Evidence drift: actual individual counts are contract `44`, protocol `125`; `git diff --check` reports seven whitespace errors; the fixture says lifecycle allows four reads while its default allowlist is empty.

One Commit-1-only corrective tranche has been sent to the owner. It requires long-lived real WS, pre-bearer key validation/rollback, strict env key-set, safe bearer serialization/terminal cleanup, lease-only API, restored Phase-1 upstream rejection, production hard-pin to approval=never, clean reproducible Node + true-PTY evidence, and corrected docs/counts. The locked Commit-2 stop/abort/timeout work must not be pulled into this corrective commit. No merge/deploy authorization.

vansin (commented):
## Independent review — fd2399c: FAIL (Commit 2 remains locked)

I reproduced the reported positive evidence from a detached checkout: frozen hashes match, the gateway subtree is **287/0/814**, and after manually generating the missing integration bundle the real-Node suite is **30/30** and the Codex 0.144.0 PTY bootstrap smoke is **4/4**. The previous “owner dies after the header timeout” regression, exact child-env allowlist, and removal of the public no-lease consume method are real fixes.

The checkpoint still fails on the following boundaries:

### P0

1. **Real upstream reverse requests are still silently dropped.** `uds-server.ts:497-500` returns for every non-response frame, and `TuiWsServer` has no upstream frame/write seam. A live fake transport emitting `approval/request` produced `writes=[]`, `diagnostics=[]`. Coordinator-only tests do not prove the Phase-1 `approval_mode_never` response reaches the real upstream with the original id.

2. **Rejected half-open sockets escape admission accounting and can block shutdown.** `onUpgrade` removes the socket from `preAuthTimers` before structural/bearer validation. `writeGenericReject` only calls `socket.end(payload)`, despite its “destroy” comment. A real Node `allowHalfOpen:true` client remained open past 3× the timeout with `preAuth=0`; `server.stop()` then did not resolve.

3. **Pending bearer/redactor bytes remain observable.** `TuiBearer` stores plaintext/digest in discoverable, writable Symbol properties; `util.inspect(...,{showHidden:true})` prints the pending plaintext, and replacing the digest Symbol permits a caller-selected presentation. The existing test calls `takePlaintextForLauncher()` before hidden inspection, so it inspects an already-cleared buffer. `SecretRedactor.secretBuf` is an enumerable TS-private field: ordinary `JSON.stringify`/`util.inspect` exposes its full byte array, and `finish()` does not clear it.

### P1

4. Duplicate `Host` headers are not counted from `rawHeaders`; a request with the correct Host first and a second external Host received `101 Switching Protocols` and held the owner slot.

5. Rollback paths call `ws.close(1006, ...)` at `tui-ws-server.ts:333,340`. 1006 is reserved and `ws@8.21.0` throws; the exception is swallowed, so the upgraded socket may remain unwired/open.

6. `maxPayload`, `headerTimeoutMs`, and `maxPreAuthSockets` are still production-facing options. The claimed 1 MiB / 3 s / 8-socket hard pins are only defaults.

7. Lease material is still public to in-process callers through `GatewayLifecycle.humanOwnerCoordinator()` plus `HumanOwnerCoordinator.currentLease()`. Prefer attach returning a typed success result and keep the raw lease out of the public lifecycle surface.

### Evidence / accuracy

8. In a clean checkout, `bun run build` emits only `dist/cli.js`. Both reported Node scripts then fail with `ERR_MODULE_NOT_FOUND: dist/rfc030-integration.mjs`. The 30/30 and 4/4 results require an unreported manual `bun build ...integration-entry.ts` step and therefore are not yet clean-reproducible.

9. The CLI harness is a bootstrap smoke (it SIGKILLs the interactive CLI), not a successful CLI lifecycle run; it observes `account/read`, not all four documented reads. Missing/wrong Codex or PTY currently exits 0 as `PASS: skipped`, which cannot be a release gate. The fixture also says lifecycle uses the four-read fake allowlist, but lifecycle is intentionally default-deny with an empty set.

10. There is no tracked `bun.lock`; direct pins resolve correctly in a clean install, but the report’s “bun.lock confirmed” statement is not reproducible. Full Commit-1 `git diff --check` also still has seven whitespace hits in `lifecycle.test.ts`.

### Required corrective (Commit 1 only)

- Add one real upstream frame router/write seam: responses retain single-mux routing; requests go through `HumanOwnerCoordinator`; Phase 1 writes a structured `approval_mode_never` error on the original request id. Notifications must have an explicit policy/diagnostic, never silent drop. Test through the actual transport subscription and write sink.
- Keep every socket accounted until successful Upgrade or close; reject responses must force-destroy within a bounded interval. Add a real-Node allow-half-open regression proving socket close, ledger cleanup and bounded `stop()`.
- Move verifier/redactor bytes to genuinely non-observable runtime storage (ECMAScript private fields or module WeakMap), clear plaintext **and digest** on terminal transitions, make `finish()` clear redactor bytes, and add pending-state JSON/inspect/showHidden/symbol/mutation red tests.
- Enforce raw-header singleton rules (at minimum Host; apply the same strict rule to the reviewed singleton WS/auth headers), replace 1006 sends with terminate/destroy or a valid close code, and remove production limit overrides.
- Provide committed one-command clean evidence that builds its own bundle; hard-fail missing prerequisites; label the CLI check honestly as bootstrap smoke and correct the fixture/lock/diff disclosures.

No Commit-2 shutdown work, no Wave 2, no merge/deploy/latest.

vansin (commented):
## Independent review — 3e62297: FAIL (Commit 2 remains locked)

Positive evidence is real:

- frozen contract/protocol hashes match;
- full subtree **290/0/815**, all per-file counts match;
- clean `npm run test:rfc030-node-integration` builds its own bundle and passes **37/37**;
- steady-state reverse request preserves the original id and writes `NoOwner / approval_mode_never`;
- WeakMap bearer storage, singleton headers, canonical key, valid WS close behavior and half-open reject cleanup all independently pass.

The checkpoint still fails:

### P0 — lifecycle-owned upstream routing is not established

There are two transport subscribers, not one router: Backend subscribes first, then TUI. Lifecycle runs preflight before either and awaits Backend start before TUI start.

Real integration probes:

- emit reverse + close during preflight → lifecycle still reaches `running`, no upstream write;
- emit reverse between Backend and TUI subscription → no write and no diagnostic;
- emit close in that gap → lifecycle still reaches `running`;
- fully started handler count is `[2 frame, 2 close]`; each normal internal response settles but also produces the false diagnostic `upstream_response_on_ws_face`;
- a proxied-TUI response is consumed by Backend's mux path and then discarded; TUI only logs it.

The standalone TUI 37/37 test begins after full startup, so it cannot cover these races. Replace both subscriptions with one lifecycle-owned dispatcher registered before any preflight/await that can touch an already-connected transport (or buffer/fail startup until routing is ready). Response must consume once and dispatch by mux origin; request goes to HumanOwner; notification has one explicit sink; close has one terminal owner. Required red tests: preflight reverse/close, Backend→TUI gap reverse/close, handler count `1/1 → 0/0`, normal response zero false diagnostic, proxied response one-to-one delivery.

### P0/P1 — EOF credential scrub remains incomplete

`SecretRedactor.finish()` returns residual tail verbatim. Feeding the first 42 characters of the 43-character bearer yields all 42 in output; only a small final-character candidate set remains. The existing test intentionally locks this behavior. At EOF, redact a suffix that matches a prefix of the protected value instead of releasing it. Also copy retained tail bytes into owned storage: current `subarray` retains the caller Buffer backing and terminal clearing mutates caller memory.

### P1 — evidence and public surfaces

- Canonical `npm run test:rfc030-real-cli-smoke` selects the workspace-local Codex 0.133.0 and exits 1. Add an explicit auditable `RFC030_CODEX_BIN` (or equivalent) and use that exact path for both version check and spawn.
- Direct 0.144.0 smoke false-greens: 3/5 runs reached owner-held with no authorizer call and exited `3/3`. The wait loop exits on owner-held; zero calls must fail, and the first method must be exactly `account/read`, not merely any member of the four-read set.
- The fixture still says lifecycle uses the four-read allowlist; lifecycle correctly remains empty default-deny.
- Claimed hard WS limits remain externally overrideable through the public two-argument constructor and exported `_createForTest` in the production bundle. `@internal` is documentation, not a runtime boundary.
- The raw lease is still publicly reachable via `GatewayLifecycle.humanOwnerCoordinator()` + `HumanOwnerCoordinator.currentLease()`, despite the report saying P1-4 is complete.
- First attach permanently closes the listener and consumes the only bearer. After normal owner disconnect, the slot is empty but reattach gets `ECONNREFUSED`. This conflicts with the accepted owner-detach/park/re-attach contract. Keep concurrent hard-one without making the lifecycle lifetime-one-shot; wire the existing launcher seam so a fresh bearer/lease can be issued on supervised reattach. Real launcher implementation remains Wave 2.
- `buildAllowlistEnv` copies inherited allowed properties from a custom prototype. Require an ordinary/null prototype and own-property reads; return a null-prototype frozen object.

### Threat-model decision parked outside A implementation

The bearer is necessarily delivered in the Codex child environment. On this Linux host, another same-UID process can read it through `/proc/<pid>/environ`. Therefore Commit 1 must not claim protection from hostile same-UID code. Either narrow the claim to stale/accidental peers, or require Wave-2 OS isolation/non-dumpable execution with a real sibling-process gate. A must not implement the production launcher in this corrective.

No Commit-2 teardown/state-machine work, no Wave 2, no merge/deploy/latest.

vansin:
Threat-model follow-up to review 4681084684: lead decision is now frozen. The default profile trusts same-UID code; this PR must rename/remove any `same-UID impersonation defense` claim and describe the bearer as an accidental/stale-peer plus guessing defense. OS-level UID/userns/`/proc` isolation is an optional Wave-2 high-security profile, not part of this Commit-1 corrective and not a default release blocker. All other FAIL findings in the review remain unchanged.

vansin:
2026-08-13 maintenance triage: CLOSE STALE CANDIDATE / DO-NOT-MERGE; preserve branch and all evidence.

This 39-file, 17.5k-line Wave-1A snapshot is based on July state, is currently DIRTY/CONFLICTING, and its own body explicitly keeps §8 review, coordinator recheck, merge and production hard-locked. Its controlling plan PR #433 has now been closed as a superseded checkpoint document. Current canonical `docs/rfcs/RFC-030-codex-tui-bridge.md` still states the production Policy Gateway is NOT implemented and production remains blocked, so closing this PR must not be interpreted as feature completion.

The implementation/evidence remains reachable by immutable commit `703374e9` and this PR history. Any future Policy Gateway work must start as a new, current-main candidate with fresh architecture/security review rather than reviving or merging this stale snapshot. Tracking issue #428 remains OPEN. No branch deletion, product change, release, or production action.

PR #574, opened 2026-08-03
fix(opencode): correlate network replies to their originating message

vansin:
> **Authored by 通信牛.** Opened on its behalf — its node shares a codex thread saturated by long autonomous turns, so its replies hit the 600s node deadline.

## The defect

A network turn checks the session is idle, then posts. Between those two moments a human can type. The assistant reply that comes back may then belong to the human's turn — and it was being returned to the network peer as the answer to *its* task.

This is not theoretical. It cost a real user nine repeated questions today, and **three separate diagnoses of mine were wrong** (starvation, delivery failure, model loop) because nobody suspected the reply-attribution path.

## The fix

The client proposes a cryptographically random `msg_anet_*` as the POST `messageID`. The decision then reads `message.info.parentID` **from the HTTP response** — a server-honoured causal link, not a locally inferred one. A reply whose `parentID` is not our `messageID` is refused.

## Independent review

Reviewed by `grok测试员`:

| point | result |
|---|---|
| can the correlation be forged | PASS — `parentID` comes from the server response; `messageID` is a correlation token the server must honour, not a local guess |
| does it swallow human replies (over-fix) | PASS — the check runs only in the network `submit()` path; the attached human TUI never passes through it |
| behaviour when `parentID` is absent | fail-closed, and the reviewer argued for it against my stated preference |

Mutation independently reproduced: disabling the ownership check turns the race test from reject to resolve. Full suite **26 pass / 0 fail / 101 assert**.

**BLOCKER: none. MAJOR: none.**

### On fail-closed

I had leaned fail-open — losing a human's message seemed worse than an occasional misdelivery. The reviewer disagreed with a better argument: on the network return path, **misattributing a human's answer to a peer is a disclosure**, while a failed task is merely retried, and the human's own TUI display is untouched either way. Accepted.

## Merge condition

The suite proves the logic against a fake server contract pinned at 1.18.1. It does **not** prove that the real 1.18.1 binary always populates `parentID` on assistant messages.

That matters precisely *because* the behaviour is fail-closed: if the field is ever absent in practice, **every network turn fails**. Before merging, probe the real binary — the reviewer flagged this and I agree it should not be waved through on a contract test. This repository has been bitten before by freezing a protocol against an assumed upstream shape instead of an observed one.

`opencode-ai@1.18.1` is installed on this host, so the probe is cheap.

## Scope

**6 files, +109/-2**, based on current `main`.

### DIFF
--- a/agent-node/src/runtime/opencode-copresence/runtime.test.ts
+++ b/agent-node/src/runtime/opencode-copresence/runtime.test.ts
@@ -77,9 +77,10 @@ if (command === "serve") {
       await new Promise(r => setTimeout(r, Number(env.FAKE_TURN_MS || 25)));
       const prompt = json.parts?.[0]?.text || "";
       const reply = "FAKE_REPLY:" + prompt;
-      messages[id].push({ info:{role:"assistant"}, parts:[{type:"text",text:reply}] });
+      const parentID = env.FAKE_RACE_HUMAN === "1" ? "msg_human_race" : json.messageID;
+      messages[id].push({ info:{role:"assistant",parentID}, parts:[{type:"text",text:reply}] });
       delete statuses[id];
-      return send(res, { parts:[{type:"text",text:reply}] });
+      return send(res, { info:{role:"assistant",parentID}, parts:[{type:"text",text:reply}] });
     }
     res.writeHead(404); res.end("not found");
   });
@@ -287,6 +288,26 @@ describe("OpenCode native serve+attach copresence", () => {
       const result = await runtime.submit("after-human", 5_000);
       expect(Date.now() - started).toBeGreaterThanOrEqual(300);
       expect(result.replyText).toBe("FAKE_REPLY:after-human");
+    } finally {
+      await runtime?.close();
+      f.close();
+    }
+  }, 15_000);
+
+  test("refuses a reply owned by a human turn that won the idle-to-submit race", async () => {
+    const f = fixture({ FAKE_RACE_HUMAN: "1" });
+    let runtime: Awaited<ReturnType<typeof openVettedOpenCodeCopresence>> | undefined;
+    try {
+      runtime = await openVettedOpenCodeCopresence({
+        binary: f.binary,
+        env: f.env,
+        cwd: f.root,
+        workDir: f.root,
+        model: "opencode/fake",
+        startupTimeoutMs: 5_000,
+      });
+      await expect(runtime.submit("network-must-own-its-reply", 5_000))
+        .rejects.toThrow("not owned by the submitted network message");
     } finally {
       await runtime?.close();
       f.close();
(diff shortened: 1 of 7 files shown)


vansin:
## Hold — this PR points at a superseded revision

The merge condition on this PR was: probe the real `opencode-ai@1.18.1` rather than trust the contract test.

**The probe refuted the candidate.**

`msg_anet_*` does not satisfy 1.18.1's ascending message-ID rule. The second network turn therefore returns the *previous* assistant message and leaves the new user message unanswered. The `parentID` guard did its job — it refused the mismatched reply — but the net effect was that every network turn after the first one failed.

So the reviewed revision was correct about attribution and wrong about the wire contract, and the contract test could not see it because we wrote the fake server ourselves. It answered with whatever we asked for.

New revision `925a2a5e` generates IDs in the official 1.18.1 format (`msg_` + 12 hex time/counter + 14 base62) and keeps the exact `parentID` attribution. Reported evidence: 27/0/103 focused, real `opencode-ai@1.18.1` `test227` with 34 unit and 19 live checks, three consecutive submits.

Independent re-review is routed. This PR stays held until that lands, and the three points being checked are: whether the ID encoding genuinely preserves ordering (same-millisecond bursts, counter reset across restarts), whether both mutations — including `use-unordered-message-id`, which pins precisely the property that broke — go red under an independent run, and whether the "live" checks really reach the 1.18.1 binary rather than another fake.

Recording this because the lesson generalises: a green contract test against a server we authored proves the code agrees with our assumptions, not that it agrees with the upstream.

vansin:
## Updated to `925a2a5e` (branch head `e15e0b4c`)

Independent re-review is in. Verdict: mergeable, with one evidence gap the reviewer declined to paper over.

**ID ordering — PASS.** `createOpenCodeAscendingMessageId` builds `ts*0x1000 + counter` into a 6-byte hex prefix plus 14 base62. Independently property-checked: 100 same-millisecond emissions are strictly monotonic, cross-millisecond ordering holds, and a 4096th emission inside one millisecond throws rather than wrapping.

**Both mutations independently red.** `drop-reply-ownership` (expected reject, received resolve) and `use-unordered-message-id` — the second pins precisely the property that broke the previous revision. Focused suite 27/0/103.

**Live evidence — partially reproduced, and the reviewer said so plainly.** It could not rebuild the real-binary green run: the host is at 96% with no `anet-test227` image present. What it *did* find were residual failure reports from the earlier run containing genuine OpenCode responses (`providerID=opencode`, assistant `msg_fc54…`, `parentID=msg_anet_*`) — the out-of-order-ID diagnosis itself.

That is worth stating precisely: **the failure artifacts prove the harness reaches the real 1.18.1 binary.** What has not been independently reproduced is the *success* path. The author reports 34 unit and 19 live checks green across three consecutive submits; nobody else has re-run it.

Known minors, recorded: a wall-clock step backwards would regress the ID prefix, and separate processes emitting in the same millisecond keep independent counters — the same shape as a human TUI writing concurrently.

Merging is held on freeing enough disk for one `test227` rebuild, not on any code finding. Everything on the code side — ordering, ownership, both mutations — has been independently verified.

PR #576, opened 2026-08-03
fix(hub): make Dashboard chat sends idempotent

vansin:
## What changed

- Accept an optional bounded `meta.client_request_id` on MCP `send_task`.
- Derive a durable task ID from authenticated network, authenticated sender, and request ID.
- Return the existing task on an exact replay after a lost response.
- Fail closed with `idempotency_conflict` if the same key is reused with a changed payload.

## Why

Dashboard retries after an ambiguous transport failure previously risked creating a duplicate task. The Hub now provides the durable idempotency boundary required by the matching Dashboard recovery flow.

## Compatibility

- No schema migration.
- `/api/task` is unchanged.
- Callers that omit `client_request_id`, including old `atok_` clients, retain the UUID path.
- No historical data cleanup or deployment is included.

## Validation

- Docker focused suite: 6 passed, 0 failed.
- Author-run full Hub suite: 786 passed, 10 skipped, 0 failed, 2264 assertions.
- Witnessed-red mutation replacing the stable ID with a random ID breaks replay, conflict, and determinism tests.
- Independent review source-audited the behavior; the shared-host disk limit prevented an independent full-suite rebuild.

Companion Dashboard Draft PR carries the UI retry, outbox, timeout, and privacy lifecycle changes.

### DIFF
--- a/docs/tests/report-test583-dashboard-chat-idempotency.txt
+++ b/docs/tests/report-test583-dashboard-chat-idempotency.txt
@@ -1,0 +1,29 @@
+Test 583 — Dashboard chat idempotency
+Date: 2026-08-03 (Asia/Shanghai)
+Base: origin/main 9c5082162284cc24850509cab58ff88a91ebfca9
+Scope: MCP send_task only; existing callers without meta.client_request_id are unchanged.
+
+Contract
+- A valid dashboard client_request_id is scoped by authenticated network and sender.
+- An exact replay returns the original task/message id and current task status.
+- Reusing the id with a different target, content, priority, or metadata fails closed
+  with idempotency_conflict.
+- The identifier is bounded and is not a credential.
+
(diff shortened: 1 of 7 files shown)


vansin:
Closing as a duplicate — this content is already on `main`.

It was reviewed, opened as #575 and merged as `4da56c34` before this PR was created. `idempotentTaskId` is present on `main` (2 call sites, verified against the merged tree).

No action needed from the author. This duplicate exists only because the coordination channel to its node is currently consuming roughly one inbox item per 600s deadline, so the merge confirmation is still queued behind other traffic. The companion Dashboard change is likewise already merged as sleep2agi/agent-network-dashboard#82.

The `review/dashboard-chat-ha-core` branch can be deleted whenever convenient; nothing depends on it.

PR #577, opened 2026-08-03
fix(codex-app-server): deliver Dashboard chat into the active human turn

vansin:
> **Authored by 通信牛.** Opened on its behalf — its bridge processes roughly one inbox item per 600s deadline.

## The defect

When a human turn is active on a shared codex thread, the bridge could only queue network tasks FIFO. Each queued item then waited out its own 600s node deadline, so **one long human turn wedged everything behind it indefinitely**. Production evidence: inbox rows with `target=通信牛, acked=0` sitting undelivered while the bridge had already received the SSE.

This is why Dashboard messages to a copresence node appeared to vanish.

## The fix

Use the app-server's `turn/steer` to append a Dashboard chat message into the turn already running, instead of waiting for idle.

**Only messages the Hub itself authenticated as coming from a user are eligible.** That constraint is what makes this safe, and it was not the first design.

## How the safety argument was reached

The first version proposed steering *any* network task into a human's turn and mapping the turn's final answer back as the reply. That is the same shape as a defect fixed earlier today (#574): a reply belonging to the human's turn returned to a network peer. Steering makes it worse, because it actively injects into someone else's conversation.

The author's answer was better than the check I asked for: **narrow the eligible set so the attribution problem cannot arise.** Only a `dashboard-chat` task stamped `auth_origin=user` by the server steers — so both sides of that turn are the same principal. Ordinary agent tasks still queue FIFO and never touch a human turn.

That moves the entire safety argument onto one assumption: **`auth_origin` is stamped server-side and cannot be forged.** So that is what the review had to pin.

## Independent review

| point | result |
|---|---|
| `auth_origin` stamped server-side on both transports; a node token supplying `auth_origin:user` is stored as `node` | PASS, mutation red on MCP and REST separately |
| no third write path — only two `INSERT INTO tasks` sites carry `meta_json` | PASS |
| ordinary agent tasks still FIFO (no accidental concurrency) | PASS |
| steer race requeues rather than drops | PASS |
| ack happens only after a reliable reply — `task_steered` is not completion | PASS |
| rolling idempotency ignores the server-only stamp without collapsing genuinely different requests | PASS |

## A legacy path that was removed rather than guarded

Review flagged a residual: a historical row with no stamp but `from_session=admin` would still qualify. Chasing that produced a harder fact — **the production Hub is `0.9.0-preview.26`, which predates the REST identity binding merged today (#571)**. On that Hub a node token can still claim `from=admin` over REST. Deploying this bridge ahead of the Hub would therefore have handed any node write access into a human's turn.

The author deleted the legacy path outright instead of hardening it. Unstamped rows now take the ordinary FIFO route; no identity is inferred from an alias.

Verified: `isInteractiveDashboardTask` requires `auth_origin === "user"`, and the legacy admin branch is gone.

## Deployment

The feature is **inert until the Hub is upgraded** — with no stamp, nothing steers. Confirmed fail-closed rather than permanently disabled: once the Hub stamps, steering activates with no additional switch.

Production currently runs Dashboard `54150269` and Hub `preview.26`. Hub first, then the bridge.

### DIFF
--- a/agent-node/src/cli.ts
+++ b/agent-node/src/cli.ts
@@ -95,6 +95,7 @@ import {
 } from "./runtime/opencode-acp/profile-state";
 import { OPENCODE_DEFAULT_PIN } from "./runtime/opencode-acp/binary";
 import { createInboxDrainLane, drainInboxBatch } from "./runtime/inbox-drain-lane";
+import { dispatchInboxBatch, isInteractiveDashboardTask } from "./inbox-dispatch";
 import { createSingleFlight } from "./util/single-flight";
 
 const home = homedir();
@@ -2890,7 +2891,12 @@ async function closeOpencodeRuntime(reason: string): Promise<void> {
 // exactly like every other runtime — the bridge only wraps "run one turn".
 // A second concurrent task queues FIFO inside the bridge and is drained when
 // the in-flight turn (ours OR a human TUI's) completes.
-async function processWithCodexAppServer(task: string, _from: string, taskId: string | null): Promise<string> {
+async function processWithCodexAppServer(
+  task: string,
+  _from: string,
+  taskId: string | null,
+  steerIfExternalTurn = false,
+): Promise<string> {
   const { openCodexAppServerRuntime, codexAppServerThink } =
     await import("./runtime/codex-app-server/runtime");
 
@@ -2936,6 +2942,7 @@ async function processWithCodexAppServer(task: string, _from: string, taskId: st
     taskId: taskId || `local-${Date.now()}`,
     text: task,
     from: _from,
+    steerIfExternalTurn,
     log,
   });
 
@@ -3297,7 +3304,13 @@ let thinkQueue = Promise.resolve();
 // OLD thinkQueue ref, and the new task is killed by exit(75) mid-flight.
 let configApplyDraining = false;
 
-function think(task: string, from: string, taskId: string | null, images?: string[]): Promise<string> {
+function think(
+  task: string,
+  from: string,
+  taskId: string | null,
+  images?: string[],
+  steerIfExternalTurn = false,
+): Promise<string> {
   if (configApplyDraining) {
     // Don't accept new work during a restart drain. The error string
     // is intentionally explicit so the upstream caller (inbox handler,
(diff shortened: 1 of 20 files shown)


vansin:
## Final revision `9f264840`, and a design claim the author retracted

The reconnect path went through three revisions, and the last one is the most interesting: the author **probed the real production app-server** (codex 0.133) with a read-only `thread/read`, and found that an active turn's history returns `agentMessage` / `tool` / `fileChange` items but **no `userMessage` at all**.

That falsifies the premise of the previous revision. The prefix heuristic cannot be applied to an in-flight turn on reconnect, because there is nothing to apply it to. The author retracted the claim that a current turn could be recovered as steerable, and pinned the real wire shape as a negative test: *missing `userMessage` → no `turn/steer`*.

Practical consequence: after a bridge restart, an in-flight human turn is FIFO-only. Steering resumes on the next live `turn/started`. Nothing is lost — queued Dashboard rows process normally once the turn ends.

This is the third time today that a real-binary probe overturned something a contract test had already passed. It is worth naming as a pattern: **a green test against a fixture we authored proves the code agrees with our assumptions, not with the upstream.**

## Residual, deliberately recorded

The prefix check is a heuristic, not an authoritative stamp:

```js
steerable = typeof firstUserText === "string" && !/^\[Agent Network(?:\/|\])/u.test(firstUserText)
```

- absent or non-string first message → **not** steerable (fail-closed)
- carries the prefix → **not** steerable, FIFO-only
- **any other text → treated as human, steerable**

Safety therefore rests on a fact about this product, not on a protocol guarantee: **every network turn this bridge opens carries `[Agent Network/…]`**, and has since Phase0A. A prefix-less network turn from some older or external implementation would be misread as human. There is no third turn-opening path in the tree today, and no authoritative "this turn is human" marker exists to fail closed against — defaulting to non-steerable would disable the feature entirely.

**Follow-up, not blocking:** the regex anchors at the literal `[`, so leading whitespace would escape it (`"  [Agent Network/…"`). This bridge never emits leading whitespace, and on the real wire the reconnect path never reaches this check at all, so the gap is currently unreachable. Worth tightening to `/^\s*\[Agent Network(?:\/|\])/u` when someone next touches this file — I asked for it, then relaxed it once the real-wire probe showed the path was dead. Recording the reasoning so the relaxation is auditable rather than silent.

## Deployment order

Production runs Dashboard `54150269` and Hub `preview.26`. **Hub first, then the bridge** — the Hub carries the REST identity binding (#571), the idempotency (#575) and the `auth_origin` stamp this change depends on. With no stamp, nothing steers, so a bridge-first deploy is inert rather than dangerous, but there is no reason to sequence it that way.

PR #607, opened 2026-08-09
feat(server): expose upload limits via health

### LINKED ISSUE #496: Expose upload limit (MAX_UPLOAD_BYTES) via /health so clients can read it authoritatively
## Context

Discovered while shipping dashboard PR #45 (#492 upload half). The dashboard mirrors hub's `MAX_UPLOAD_BYTES` (12 MiB) as a hard-coded constant to do a client-side pre-check — otherwise an oversize request races the hub's early 413 close and surfaces to callers as `fetch failed` (502) instead of a real 413 with a Chinese human message.

Two hard-coded constants for the same rule = silent drift risk. 通信龙 flagged this in the #45 pre-review, and noted a real blind spot in a purely client-side runtime drift-detect:

- **L_dashboard > L_hub**: files between the two caps get rejected at hub, dashboard sees `limit_bytes` in hub's 413 body → detectable ✅
- **L_dashboard < L_hub**: files between the two caps are rejected client-side, hub is never contacted → **silently undetectable** 🔴

The second case is the more damaging one — legitimate files get rejected with "超过 12 MB 上限" while the hub would actually accept them, and nothing logs the discrepancy.

## Proposal

Add `max_upload_bytes` (and `max_request_content_length`) to the hub's `/health` response, OR expose them at a dedicated `/api/limits` endpoint. This becomes the single source of truth; the dashboard can then read it at startup, cache the value, and use it for pre-check.

## Requirements

- Endpoint MUST be unauthenticated (or use lightweight cache) — dashboard startup pings shouldn't burn auth cycles
- Response must be stable across hub versions (add fields, don't rename)
- Dashboard-side (separate PR) fetches once at boot, keeps a fallback to its hard-coded mirror value if fetch fails, **loudly logs when the fallback is in use** so silent fetch failure doesn't silently re-introduce two constants

## Placement suggestion

`/health` seems natural (already carries version + capability fields):

```json
{
  "ok": true,
  "version": "0.9.x",
  ...
  "limits": {
    "max_upload_bytes": 12582912,
    "max_request_content_length": 13631488
  }
}
```

## Non-goals

- Shared npm package for the constant — dashboard and hub are independently released, coupling them via a shared package creates a new drift mode ("upgraded one didn't upgrade the other").

## Related

- PR #45 (dashboard upload proxy) — ships with a comment naming this issue as the follow-up
- Pre-review conversation with 通信龙 — flagged the blind spot

vansin:
## Summary

- expose Hub-enforced upload and request-size limits in the anonymous `/health` response
- source both values from the existing enforcement constants
- preserve the existing health contract through an additive `limits` object
- add real HTTP coverage and an exact witnessed-red mutation

## Verification

- Docker test607 exact source: `d788d2083967bc57a60f93461b1cb4237c070381`
- Hub build: PASS
- real anonymous HTTP tests: 2 tests / 9 assertions
- exact health-binding mutation: FAIL as required, restored source PASS
- report: `docs/tests/report-test607-health-upload-limits.txt`

Fixes #496

### DIFF
--- a/docs/tests/report-test607-health-upload-limits.txt
+++ b/docs/tests/report-test607-health-upload-limits.txt
@@ -1,0 +1,32 @@
+Test 607 — Hub authoritative upload limits
+
+Issue: https://github.com/sleep2agi/agent-network/issues/496
+Base commit: 5b2688fdc41e8c5bfa57820645abcf1f09ffe3df
+Source commit: a90c0cf64231702f2c06dd309471653314d579a6
+
+Scope
+- Adds an anonymous, additive `limits` object to `GET /health`.
+- Values are sourced from the same server constants that enforce upload and request limits.
+- Existing health fields and authentication behavior are unchanged.
+
+Docker evidence
+- Image: anet-test607-hub:dev
+- Image ID: sha256:b34d34c6b768ee280dedf5e719e7a8e445fe7d1997b461c0fc8aa8e53a39d895
+- Embedded TEST607_SOURCE_COMMIT: a90c0cf64231702f2c06dd309471653314d579a6
+- Artifact SHA256: 4c792b5a4d0fe6b20cb8241f064130e914df38362e5cbc1c799de49134b3c656
+
+Results
+- Real Hub build: PASS (257 modules, 1.52 MB output)
+- Real HTTP health tests: PASS (2 tests, 9 assertions)
+- Anonymous health response retains watchdog-compatible shape: PASS
+- `max_upload_bytes` equals the enforced Hub upload constant: PASS
+- `max_request_content_length` equals the enforced request constant: PASS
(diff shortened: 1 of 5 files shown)


vansin:
✅ rebase 窄 provenance 复核 MERGE:rebase 到当前 main 5b2688fd 干净(diff 仅 health-upload-limits:server.ts +8 additive /health limits.max_upload_bytes、test、docker、report,无混入);fresh SOURCE_COMMIT=a90c0cf6 匹配,2/0/9,max_upload_bytes:MAX_UPLOAD_BYTES→1 binding mutation rc1(报告 claim 现与 run.sh 一致,修正了之前 #496 mutation 计数误报)。叠加原 review-PASS,MERGE。server 合 main、部署等 Vincent GO。

PR #701, opened 2026-08-12
fix(#698): terminalize peer replies across compatibility paths

### LINKED ISSUE #698: fix: codex app-server peer replies create new tasks without terminalizing originals
## Symptom

Codex app-server family nodes successfully produce and send replies, but the original inbound task remains `running`/`acked` forever. Each reply is persisted as a brand-new task instead of closing the task being answered.

Field evidence on an online Windows `codex-app-server-sdk` node:

- the node log says it is sending a reply for the original task with `status=replied`;
- the recipient receives the reply under a different task ID;
- the original task stays non-terminal;
- repeating ordinary checks makes `running`/`acked` counts grow monotonically even while the process is idle and `process_in_flight_count=0`.

This is distinct from #696: #696 drops short reply text before transport; this issue occurs even when a substantive reply is successfully transported.

## Root cause (origin/main)

`agent-node/src/cli.ts` deliberately sets:

```ts
const REPLY_VIA_SEND_TASK = RUNTIME === "codex-app-server";
```

For a routable agent peer, `sendReply()` then calls `send_task` with `parent_task_id=taskId` instead of `send_reply`.

That preserves the RFC-030 “wake the peer immediately” behavior, but loses task lifecycle closure:

- Hub `send_reply` atomically updates the original `tasks` row to `replied|failed|cancelled`, writes `result/completed_at`, records the event, and synchronizes a bound scheduled run.
- Hub `send_task(parent_task_id=...)` creates another `tasks` row. Parent chaining happens only after that child later reaches a terminal state; it does not terminalize the task that the node is currently answering.
- The newly created reply task itself requires processing, which can create reply ping-pong and additional non-terminal rows.

The source comments currently frame the choice as “peer wake **or** lifecycle close”; the protocol needs both.

## Required invariant

A final agent-to-agent reply must, as one logical operation:

1. terminalize the exact original task using stable `task_id` and network scope;
2. persist the final result/status and `completed_at`;
3. synchronize any bound `scheduled_task_runs` terminal state;
4. wake the original peer immediately with exactly one actionable reply;
5. not create a new requires-response task and not start a reply ping-pong loop.

Dashboard/API-originated replies must retain their existing behavior.

Do not “fix” this by canceling old rows, hiding counts in the UI, or invoking `send_reply` followed by `send_task` as two independent writes. Those approaches either treat symptoms or permit half-state/duplicate delivery on failure.

A valid design may add an atomic Hub peer-reply primitive, or make the existing reply primitive provide peer wake semantics without creating a second task. The implementation choice is open, but the invariant above is not.

## Acceptance tests

Run in Docker with exact-source report and witnessed-red mutations.

### Core lifecycle

- Agent A dispatches task T to a Codex app-server node B.
- B returns a successful final response through the real agent-node route.
- T becomes `replied` exactly once with the exact result and non-null `completed_at`.
- A receives exactly one immediate actionable reply.
- No new requires-response task is created solely to carry B's reply.
- Failure/cancel statuses terminalize T with the matching status.

### No ping-pong / bounded rows

- Run at least three A↔B exchanges.
- Each original task reaches terminal state.
- Task-row growth equals dispatched work only; replies do not recursively create work tasks.
- `running`/`acked` returns to zero after completion.
- Replaying the same final reply is idempotent/no-op or returns a precise already-terminal error; it never delivers twice.

### Safety and compatibility

- network isolation is enforced on the original task lookup and peer delivery;
- node-token identity cannot terminalize another node's task;
- stale/reassigned task ownership is rejected;
- terminal update + reply delivery are atomic under injected failure (no terminal-without-delivery and no delivery-without-terminal);
- bound scheduled-task runs reach the same final status in the same transaction;
- Dashboard/non-session sender path remains unchanged;
- Claude/Grok/OpenCode existing reply behavior does not regress;
- exact runtime spellings/normalization for `codex-app-server` and observed `codex-app-server-sdk` are covered.

### Mutation gates

At minimum, witnessed-red mutations must prove:

- removing original-task terminalization;
- reintroducing a reply `send_task` that creates a requires-response task;
- removing peer wake delivery;
- splitting the atomic transaction;
- dropping network/owner scope.

## Existing-row cleanup boundary

Historical leaked `running`/`acked` rows may be cleaned only after the protocol fix, by a separate owner-controlled, auditable migration. This issue does not authorize broad production DB updates or fleet restarts.

vansin:
## Superseded

Superseded by merged PR #703 (`be1c7dbb`; current main includes it). Frozen report commit `791d3394` is itself immutable and remains reachable in this branch history. The branch was later fast-forwarded by mistake with `63f64a7a` and `79a1faf0`; those appends did not alter `791d3394`, but they mean the branch tip is no longer that historical report coordinate. The authoritative implementation and reproducible evidence are **only PR #703**. Do not merge, force-push, rewrite, or delete this branch; it is retained as an additional reachability path for the historical commits.

## What changed

Fixes #698 by making peer replies terminalize the original task across both current and legacy Hub compatibility paths. Capability downgrade now uses terminal `send_reply`, never a response-requiring `send_task`; SSE is only a doorbell, not a capability requirement. Current-Hub compatibility replies are token/alias bound to the exact assignee and immutable original sender while still allowing legitimate node-id rotation.

## Frozen coordinates

- Base: `17b8223f9d7fd25fcc435b40be8402f4d49df1087391ae`
- Source: `69ad300884bc91d82fff87234cd195aa3f3bb284`
- Report-only head: `791d3394530396a6390a868027262dccd20067f6`
- Exact Docker image: `anet-test698:69ad3008-exact`
- Image ID: `sha256:138e382803aa8df022174979ba2e3a4cc3e2cf3f3f1892f8ff2bb6317f1d6edc`
- Independent verdict: [CLEAN (5 MINOR), merge recommended](https://github.com/sleep2agi/agent-network/issues/698#issuecomment-5261500574)

## Evidence

- Exact image environment pins `TEST698_SOURCE_COMMIT=69ad300884bc91d82fff87234cd195aa3f3bb284`.
- Frozen-source byte verification passed; independent review expanded it to **375/375** repository files with zero mismatch/absence.
- Full Docker suite: **178 pass / 0 fail / 518 expect**.
- Main mutation matrix: **22/22 red**; legacy-wire matrix: **6/6 red**; no survivor or no-op.
- Archived real old-Hub binary returned the real `-32602` unknown-tool shape and passed both wire and production-cli terminal fallback probes.
- Independent review added the missing current-Hub/no-live-SSE probe: original task `replied`, exactly one reply, zero child tasks. The same probe reproduced the prior candidate's permanent non-terminal failure before passing on this source.
- Foreign closer and wrong-target attempts are rejected before writes; node-id rotation still terminalizes by token-bound alias.

## Known non-blocking MINORs

1. `getSSEStats` is now a dead production import retained only by a test mutation.
2. Two `run_*_wiring_mutation` helpers are defined but never called.
3. `run-legacy-wire.sh` reads its Dockerfile from the worktree instead of the archived commit, so a checkout without that directory fails to reproduce. Prefer materializing the Dockerfile from the frozen commit.
4. Report §4 groups five downgrade triggers in one sentence although only the archived-old-Hub unknown-tool trigger was exercised against a real old binary; the limits section is accurate, but the standalone sentence is too broad.
5. `reply-routing-source.test.ts` is a weak source-string assertion. It is non-load-bearing because the independent wrong-field and missing-`in_reply_to` behavioral mutations both turned red.

## Boundary

This draft PR does **not** authorize merge, npm publication, Hub deployment, fleet restart, or production rollout. Merge and any TM微软牛 pilot remain separately gated on Vincent's explicit decision and the single-node rollback/UAT SOP.


vansin:
## Superseded by #703(已合并)—— 关闭

`#703` 已于 `2026-08-12T04:33:04Z` 合并进 main(`be1c7dbb`),其后 `#700` 合并为 `6e87a346`。本 PR 的修复已由 `#703` 以更窄的形式落地,**权威实现只认 `#703` 的 `e920ebf8`(source) / `c4af6676`(report)**。

### 关于 body 里那句 "791d3394 immutable historical anchor"

需要更正措辞,但**证据没有丢**:

- `791d3394` 是 commit 对象,**按构造不可变**;分支后来的追加是 **fast-forward**(非 rebase / 非 force-push),所以它**仍然在本分支历史中可达**,`git show 791d3394` 与引用它的裁定坐标照常解析。
- 漂掉的只是**分支 tip**:本分支后来误追加了 `63f64a7a`(脚本)与 `79a1faf0`(report)。原因是执行方在上下文压缩后按一条旧授权继续执行,没有先核 PR body 与 `#703` 的现状。

### 处置

- **不 force-push、不 rewrite** —— 用重操作改写历史只为让现实迁就一句描述,方向是反的;代价是可能把上述两个 commit 变成孤儿,收益只是一句话。
- 🔴 **分支保留,不删** —— 让这几个 SHA 多一条可达路径。
- 本 PR 关闭为 superseded。

### 这一轮的工作没有白费

为本分支跑的那套门,独立复现了 `#703` 的同一批结论,可作交叉验证:`safe_rm_rf` 对真实 `TMP` 的实测(allow-list 命中、目录确已删除)、主套 `178 pass / 0 fail / 518 expect`、`28/28 MUTATION_RED`、在**不含 `tests/test698-atomic-peer-reply/` 的 base worktree** 里旧 Hub runner 跑通(`LEGACY_PRISTINE_PASS` + 2/2 red)、provenance `25/25 MATCH`、CI 8/8。

执行方在发现"历史拓扑不满足窄 delta(中间夹了旧 report commit,实际是 2 文件)"后**主动停手、如实上报、没有伪称窄、没有自行 force-push** —— 这个处置是对的。

PR #877, opened 2026-08-15
fix(scheduled-tasks): POST /cancel alongside DELETE, both idempotent

vansin:
Fixes production 定时任务「取消」P0 (dispatch 通信狗 08342434). Paired with dashboard PR (link once opened).

## Root cause

Production `y.vansin.top:3100/scheduled-tasks` — clicking 取消 fires `DELETE /api/hub/scheduled-tasks/:id` through the dashboard proxy. Some reverse proxies in the delivery path rewrite DELETE to a 405 HTML response; the dashboard then called `await res.json()` on the HTML and blew up with "Unexpected token `<`", hiding the failure from the user (spinner stuck, no error surfaced).

## Fix in this PR

Add `POST /api/scheduled-tasks/:id/cancel` as an alternate spelling for the same cancel operation. POST is universally allowed; DELETE stays for API compatibility.

- **regex** (`server/src/scheduled-tasks.ts` L451): `(runs|run-now)` → `(runs|run-now|cancel)`
- **handler**: `(!sub && DELETE) || (sub === 'cancel' && POST)` → same implementation (`status=cancelled`, `next_run_at=NULL`, `revision+1`, `updated_at=now`)
- **idempotent**: an already-cancelled row now returns `200 {ok:true, status:'cancelled'}` **without** touching the DB — previously the UPDATE would bump `revision` on every retry, which races with the dashboard's `revision_conflict` handling (retries would 409 themselves).

## Tests

`server/src/scheduled-tasks-http.test.ts` gets 3 new assertions inside the existing `optimistic revision, pause/resume, run-now and cancel preserve history` test:

1. `POST /cancel` returns 200 with `status:'cancelled'` on a live row
2. `DELETE` on already-cancelled = 200 + revision **unchanged** (idempotent)
3. `POST /cancel` on already-cancelled = 200 + revision **unchanged** (idempotent)

Baseline on this branch already has 3 pre-existing failures unrelated to cancel (`non-overlap skips` data ordering + `runs.length >= 4` count flake in this same test's tail); this PR does not touch those, and adds 7 new `expect()` calls that all pass. Full run: baseline 9 pass / 3 fail / 99 expect → this PR 9 pass / 3 fail / **106 expect** (delta +7 pass).

## 🔴 Deploy ordering constraint

**Hub must ship before or at the same time as the paired Dashboard PR.** Once dashboard is on the new UI, cancel clicks hit `POST /cancel`; if hub is still on the old handler, that path 404s and cancel appears broken again (different symptom, same user pain). Recommend: land this PR first, deploy hub to `:9200`, then deploy dashboard to `:3100`.

## Paired PR

Dashboard side: (link added on open) — updates `mutate('cancel')` to POST `/cancel`, adds safe response parsing, confirm dialog, cancelled filter, and `type="button"` on all row actions.

### DIFF
--- a/docs/tests/report-test604-scheduled-task-edit.txt
+++ b/docs/tests/report-test604-scheduled-task-edit.txt
@@ -1,0 +1,78 @@
+# test604 — Hub scheduled-task editing
+
+Date: 2026-08-09 (Asia/Shanghai)
+
+## Exact source coordinates
+
+- Hub base: `0546365dad96aa1d2dca36edbe71761b48aa3ba4`
+- Hub source: `fa6a07b7dd00314e34f6729221a577194a6a80a0`
+- Dashboard base: `79723e419d8549c24dd7e1a5969f923abc241127`
+- Dashboard source: `3c4ed457d5c8f0f05f2ec396577b81ec759bd149`
+- App base: `30fbe916763b3421c61de99e5340e89638043f29`
+- App source: `79eab938229351312e46f00df80af4439f4d80c4`
+
+All three source worktrees were clean after their source commits. No production
+runtime, database, global npm installation, or existing dirty checkout was
+modified.
+
+## Behavior delivered
+
+- Dashboard and Expo App expose Edit only for active/paused schedules.
+- The form restores and can update name, stable target node, task content,
+  priority, schedule, IANA timezone, and misfire policy.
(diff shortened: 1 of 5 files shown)


vansin:
Superseded by fresh PR cherry-picked onto current main (feat/scheduled-task-edit was 2 weeks behind, contained already-merged commits). Same fix, cleaner base.

PR #1161, opened 2026-08-24
chore(docs): point the desktop updater at 0.2.31

s2agi:
Points the served desktop updater manifest at **0.2.31**.

## Source of the file

Downloaded **anonymously, with no token**, from the published release asset:

```
env -u GITHUB_TOKEN -u GH_TOKEN curl -H 'Accept: application/octet-stream' -L \
  https://api.github.com/repos/sleep2agi/agent-network-app/releases/assets/526922959
```

and written in verbatim. The served file and the release asset are now
**byte-identical** (`md5 74ff3f9222e252070f902dd032643ef1`), where previously
they were merely semantically equivalent with different formatting — that gap is
the drift risk filed as agent-network-app#125.

## Verified before writing

- `version` = `0.2.31`
- 5 platform entries — `darwin-aarch64`, `darwin-aarch64-app`, `windows-x86_64`,
  `windows-x86_64-nsis`, `windows-x86_64-msi` — **each carries a signature**
- every asset URL resolves to an asset of release `375394792`
  (`desktop-v0.2.31`): `526919324` app.tar.gz, `526922884` setup.exe,
  `526922810` msi
- `notes` is user-facing text, with none of the maintainer wording that shipped
  in 0.2.30

## Ordering

The release is already **public** (published `2026-08-24T02:11:44Z`, tag
`desktop-v0.2.31` → `054ed937bfd35fe82ab8c15d4b943d79161c1b3e`), and its assets
were confirmed to fetch anonymously (302 → 200, exact content-length). That
order matters: a draft release's assets are not anonymously downloadable, so
syncing the manifest first would hand every client an update pointing at
something it cannot fetch.

## Gate

```
$ node docs-site/scripts/check-desktop-update-route.mjs
desktop updater static manifest: 0.2.31 ok
```

按内容搜过:`0.2.31` ⇒ 零命中;`latest.json` / `desktop/update` ⇒ 命中 #1160、#1154、#1155(均已合并,分别是改直出路由、建更新端点、首页下载区,与本次同步无关)。

## Note

Merging this does not put it online — the `docs-site` Vercel project is not
connected to git. The deploy is a separate manual step, per
`deploy/docs-site/README.md`.

### DIFF
--- a/docs-site/docs/public/desktop/update/latest.json
+++ b/docs-site/docs/public/desktop/update/latest.json
@@ -1,27 +1,27 @@
 {
-  "version": "0.2.30",
-  "notes": "Signed macOS and Windows desktop update. Publish this draft only after installation smoke tests pass.",
-  "pub_date": "2026-08-23T21:15:43.139Z",
+  "version": "0.2.31",
+  "notes": "Signed and notarized desktop update for macOS (Apple Silicon) and Windows (x64). Existing installations can update in place; new installations use the assets below.",
+  "pub_date": "2026-08-24T02:07:09.112Z",
   "platforms": {
     "darwin-aarch64": {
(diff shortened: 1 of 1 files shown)


s2agi:
## Inherited failure — `doc source-pin floor (Docker)`

This check is red on this PR, and it was already red on `main` at the commit
this branch is based on. It is **not** introduced here.

|  | run | commit | failing job | message |
|---|---|---|---|---|
| main | [32676613785](https://github.com/sleep2agi/agent-network/actions/runs/32676613785) | `9ec06a55` | `doc source-pin floor (Docker)` | `FAIL: 预期扫 110 个文档文件(= git ls-files 的结果),实际 111` |
| this PR | [32682508416](https://github.com/sleep2agi/agent-network/actions/runs/32682508416) | `0ef41916` | same job | same line, same numbers |

Both runs also print the same preceding line
(`OK: 失效 pin 1 个,基线 1 条 …`), so the pin logic itself is fine; the
mismatch is the **document count baseline**, which expects 110 files while
`git ls-files` now returns 111.

This branch changes exactly one file, and modifies rather than adds it:

```
M	docs-site/docs/public/desktop/update/latest.json
```

A modification cannot change the number of tracked files, so this diff can
neither cause nor fix that count. Tracked separately — deliberately not fixed
here, to keep the manifest sync a single-file change.

PR #1164, opened 2026-08-24
chore(docs): point the desktop updater at 0.2.32

s2agi:
Points the served desktop updater manifest at **0.2.32**, which carries the fix
for `已送达 ✓` appearing under messages the viewer did not send.

## Source

Downloaded **anonymously, with no token**, from the published release asset and
written in verbatim:

```
env -u GITHUB_TOKEN -u GH_TOKEN curl -H 'Accept: application/octet-stream' -L \
  https://api.github.com/repos/sleep2agi/agent-network-app/releases/assets/527088907
```

Served file and release asset are **byte-identical**
(`md5 860758124b7ba5f7c9b25fccf140103f`), the same discipline adopted for 0.2.31.

## Verified before writing

- `version` = `0.2.32`
- 5 platform entries, **each with a signature**
- every asset URL resolves to an asset of the published `desktop-v0.2.32`
  release: `527085503` app.tar.gz, `527088814` setup.exe, `527088748` msi

## Ordering

The release is already public (`2026-08-24T04:41:10Z`, tag `desktop-v0.2.32` →
`ee43181ed6635e98584122c66f2291115a1e0959`), and its assets were confirmed to
fetch anonymously (302 → 200, exact content-length) before this sync. A draft
release's assets are not anonymously downloadable, so the reverse order would
hand every client an update pointing at something it cannot fetch.

## Gate

```
$ node docs-site/scripts/check-desktop-update-route.mjs
desktop updater static manifest: 0.2.32 ok
```

## Duplicate check

Path overlap against **all 11 open PRs** (no result limit), on
`docs-site/docs/public/desktop/update/latest.json`: **零命中** — no open PR
touches this file.

## Known inherited failure

`doc source-pin floor (Docker)` is expected to fail here as it does on `main`:
the gate expects 110 tracked doc files and finds 111, unrelated to this
single-file modification. Tracked in #1162, with both run links and the
identical failure line recorded there.

## Note

Merging this does not put it online — the `docs-site` Vercel project is not
connected to git; deployment is a separate manual step.

### DIFF
--- a/docs-site/docs/public/desktop/update/latest.json
+++ b/docs-site/docs/public/desktop/update/latest.json
@@ -1,27 +1,27 @@
 {
-  "version": "0.2.31",
+  "version": "0.2.32",
   "notes": "Signed and notarized desktop update for macOS (Apple Silicon) and Windows (x64). Existing installations can update in place; new installations use the assets below.",
-  "pub_date": "2026-08-24T02:07:09.112Z",
+  "pub_date": "2026-08-24T04:37:25.430Z",
   "platforms": {
     "darwin-aarch64": {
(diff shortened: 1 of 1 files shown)


s2agi:
## CI status after `rerun --failed`

Run [32691108306](https://github.com/sleep2agi/agent-network/actions/runs/32691108306) was re-run for its failed jobs. The only job that is not `success` or `skipped` is:

| job | conclusion | note |
|---|---|---|
| `doc source-pin floor (Docker)` | failure | inherited — red on `main` at the base commit, tracked in #1162 |

`grok suites (Docker) (test219-grok-copresence)` **passed on the re-run** after failing on the first attempt, so it was flaky rather than caused by this change. Recording that explicitly: this diff modifies one JSON file under `docs-site/docs/public/`, which has no path relationship to the grok co-presence suite, and the second run confirms it.

The inherited failure is unchanged by this diff — a single-file modification cannot alter the tracked-file count the gate measures (110 expected, 111 found).

PR #1165, opened 2026-08-24
feat(site): add mobile downloads, and stop pointing desktop at v0.2.13

s2agi:
Adds Android and iOS entries to the download area, from
[`mobile-v0.2.32`](https://github.com/sleep2agi/agent-network-app/releases/tag/mobile-v0.2.32)
— built from `8a454431d0f952296195067471cc01fecaca0920`, the same commit as
`desktop-v0.2.32`.

## Android

Primary link straight to the APK. Card states **Android 7.0+ · 76.3 MB**:

- 76,259,813 bytes — the real APK, not the 37 MB compressed artifact figure a
  run page shows;
- the build is signed with APK Signature Scheme v2 only, so Android below 7.0
  (API 24) cannot install it. Left unstated, those users get "解析包时出现问题"
  with nothing to explain it.

Anonymous reachability verified before writing the link:
`302 → 200`, `content-type: application/vnd.android.package-archive`,
`content-length: 76259813`.

## iOS — a card that does not pretend

**The iOS card is not a link.** There is no public TestFlight invite URL yet, and
a card that looks clickable but goes nowhere is worse than one that plainly says
it is not ready. It renders dashed and muted, `aria-disabled="true"`, reading
"TestFlight 处理中 / 公开测试链接即将开放".

The `.ipa` appears only under "高级：审计用 .ipa 与 SHA256SUMS", with an explicit
line that it is App Store distribution-signed and **cannot be sideloaded**.
Offering it as an install button would send people down a path that cannot work
— it has no development provisioning and no registered devices.

When a TestFlight public link exists, that card becomes the primary iOS link in a
one-line follow-up.

## Beyond the dispatch — flagged, revert if unwanted

Every desktop link on this page still pointed at **`desktop-v0.2.13`**, nineteen
releases back, advertising **11.5 MB / 7.8 MB**. The primary "下载 macOS 版"
button was shipping a build from before the local Hub was bundled — while the
updater manifest at `www.anet.sh/desktop/update/latest.json` already serves
0.2.32.

Adding mobile links beside that would have left the more visible defect sitting
in the same block I was editing, so the desktop links, sizes and version labels
now read 0.2.32 (36.5 MB / 35.2 MB, real bytes). If this should ship as its own
PR, revert that hunk — the mobile addition stands without it.

## CSS

`.mobile-download` reuses the `.desktop-download` layout rules rather than
duplicating them, and `.download-card-pending` styles the non-interactive iOS
card (dashed border, muted text, no hover lift).

## Checks

- `node docs-site/scripts/check-desktop-update-route.mjs` → `desktop updater static manifest: 0.2.32 ok`
- `npm run build` (VitePress, dead links fail the build) → clean, 35.0s
- built `dist/index.html` contains exactly the three intended asset links and the
  pending iOS copy; **zero** `0.2.13` references remain anywhere in the page

按内容搜过(路径重叠,全部 11 个 open PR,不加 --limit):`docs-site/docs/index.md` 与
`docs-site/docs/.vitepress/theme/custom.css` ⇒ **零交集**。

`doc source-pin floor (Docker)` is expected red as inherited from `main` (#1162).

### DIFF
--- a/docs-site/docs/.vitepress/theme/custom.css
+++ b/docs-site/docs/.vitepress/theme/custom.css
@@ -929,14 +929,16 @@
   transform: translateY(-1px);
 }
 
-/* ---------- Desktop download / product paths ---------- */
+/* ---------- Desktop / mobile download, product paths ---------- */
 .desktop-download,
+.mobile-download,
 .product-path {
   max-width: 1152px;
   margin: 72px auto 0;
   padding: 0 24px;
 }
-.desktop-download {
+.desktop-download,
+.mobile-download {
   padding: 42px;
   border: 1px solid var(--vp-c-divider);
   border-radius: 24px;
@@ -951,6 +953,7 @@
   letter-spacing: 0.14em;
 }
 .desktop-download h2,
+.mobile-download h2,
 .product-path h2 {
   margin: 8px 0 10px;
   border: 0;
@@ -998,6 +1001,17 @@
 .download-card-primary {
   background: linear-gradient(145deg, rgba(0, 158, 126, 0.12), var(--vp-c-bg));
 }
+/* The iOS card is deliberately not a link: TestFlight has no public invite URL
+   yet, and a card that looks clickable but goes nowhere is worse than one that
+   plainly says it is not ready. */
+.download-card-pending {
+  border-style: dashed;
+  background: var(--vp-c-bg-soft);
+  color: var(--vp-c-text-2) !important;
+  cursor: default;
+}
+.download-card-pending:hover { transform: none; border-color: var(--vp-c-divider); box-shadow: none; }
+.download-card-pending strong { color: var(--vp-c-text-2); }
 .download-platform { color: var(--vp-c-text-2); font-size: 0.78rem; }
 .download-card strong { margin-top: 7px; font-size: 1.18rem; }
 .download-card small { margin-top: auto; color: var(--vp-c-text-2); }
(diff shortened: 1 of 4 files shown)


s2agi:
## Why `qa-frontdoor-docs` went red, and what it revealed

The first push reddened `L0 + L1 (report-only)` on `qa-frontdoor-docs`:

```
- Chinese rendered homepage lost the stable desktop release
- Chinese rendered homepage lost the macOS download
- Chinese rendered homepage lost the Windows download
```

That failure was mine, and it was correct: the guard asserted the rendered page
contains the literal `desktop-v0.2.13`.

Two things fall out of it.

**1. The guard was pinning the stale state.** "The homepage must offer a desktop
download" had become "the homepage must offer 0.2.13 forever" — so every attempt
to update reddened, and the page sat nineteen releases behind while
`anet.sh/desktop/update/latest.json` already served 0.2.32 to the in-app updater.
A gate that reds on the correct change protects staleness, not availability. The
assertions now match the shape (a desktop release link, a `.dmg`, a `.exe`) and
leave the version free.

**2. Only the Chinese page failed — because I had only updated the Chinese page.**
The English homepage was still on 0.2.13. The literal pin could not express the
invariant that actually matters here, so it is now added explicitly: **both
languages must advertise the same desktop version**. The English page is updated
in this push, with the same 0.2.32 links and the mobile section.

## Evidence

The suite was run in its own Docker image before pushing:

```
docker build -f tests/qa-frontdoor-docs/Dockerfile --build-arg SOURCE_COMMIT=<head> -t anet-frontdoor:dev .
docker run --rm anet-frontdoor:dev
→ PASS: VitePress production build
→ PASS: rendered bilingual homepages contain only the concise front door
→ exit 0
```

**Witnessed-red on the new invariant** — the English page alone reverted to
0.2.13, i.e. exactly the half-update this branch made a moment ago:

```
FAIL (1)
- homepages advertise different desktop versions (zh=0.2.32, en=0.2.13)
exit 1
```

`doc source-pin floor (Docker)` remains red as inherited from `main` (#1162).

PR #1166, opened 2026-08-24
chore(desktop): offer signed beta 0.2.33-1 via updater

vansin:
## Summary

- point the desktop updater manifest at the signed `0.2.33-1` beta
- allow numeric-only prerelease identifiers in the static route guard
- keep stable versions valid while rejecting MSI-incompatible identifiers such as `beta.1`

This intentionally makes in-app **Check for Updates** offer the signed beta to clients currently on the stable channel.

## Verification

- `node docs-site/scripts/check-desktop-update-route.mjs`
- `npm ci`
- `npm run build`
- manifest references assets from public prerelease `desktop-v0.2.33-1`

### DIFF
--- a/docs-site/docs/public/desktop/update/latest.json
+++ b/docs-site/docs/public/desktop/update/latest.json
@@ -1,27 +1,27 @@
 {
-  "version": "0.2.32",
+  "version": "0.2.33-1",
   "notes": "Signed and notarized desktop update for macOS (Apple Silicon) and Windows (x64). Existing installations can update in place; new installations use the assets below.",
-  "pub_date": "2026-08-24T04:37:25.430Z",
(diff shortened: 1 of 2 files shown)


vansin:
Updater-specific verification is green: `check-desktop-update-route.mjs`, `docs-site-build`, and a clean local `npm ci && npm run build` all pass. The remaining failures are outside this two-file docs-site change: (1) `doc source-pin floor` is the existing main failure; (2) test219 fails in its pinned runtime integration (`source_commit=485b312...`) and does not read either changed file; (3) test225 failed because a headless node process did not stop within 60s and likewise does not read either changed file. The manifest change itself references the already-public signed `desktop-v0.2.33-1` assets.

PR #1304, opened 2026-08-27
Add desktop user push and inbox count

vansin:
## Summary
- add user-keyed Desktop/web SSE stream at `/events/users/me` using `\0user:<networkId>:<userId>` keys
- add `send_desktop_message` MCP tool with same-network authorization, fail-closed target resolution, audit logging, and multi-client fanout
- keep legacy `/events/:alias` and `/events/network/:network_id` paths live and isolated
- fix client unread count path by making `/api/messages?alias=...` actually filter by alias and return `pending_count`, then render the count in Chat

## Review gates covered
- cross-network auth is tested for user stream and `send_desktop_message`
- unknown/missing/conflicting desktop targets fail closed before audit/push
- fanout test reads real SSE bytes from two live user subscriptions and verifies one cancelled client does not break the other

## Test
- `sg docker -c 'docker build --build-arg SOURCE_COMMIT=$(git rev-parse HEAD) -t anet-test1280-desktop-user-push -f tests/test1280-desktop-user-push/Dockerfile . && docker run --rm anet-test1280-desktop-user-push'`\n- Result: 33 pass / 0 fail across server tests; `anet-client-app` production build passed\n\nReport: `docs/tests/report-test-desktop-user-push.txt`

### DIFF
--- a/.github/workflows/qa.yml
+++ b/.github/workflows/qa.yml
@@ -127,6 +127,7 @@ on:
       - 'tests/test671-task-runtime-evidence-hub/**'
       - 'tests/test688-owner-gated-schedule-hub/**'
       - 'tests/test629-mcp-schema-policy/**'
+      - 'tests/test1280-desktop-user-push/**'
       - 'tests/qa-anet-daemon-cmd/**'
       - 'tests/qa-hub-16-rest-task-api/**'
       - 'tests/qa-node-04-codex-app-server-reply-routing/**'
@@ -287,6 +288,7 @@ on:
       - 'tests/test671-task-runtime-evidence-hub/**'
       - 'tests/test688-owner-gated-schedule-hub/**'
       - 'tests/test629-mcp-schema-policy/**'
+      - 'tests/test1280-desktop-user-push/**'
       - 'tests/qa-anet-daemon-cmd/**'
       - 'tests/qa-hub-16-rest-task-api/**'
       - 'tests/qa-node-04-codex-app-server-reply-routing/**'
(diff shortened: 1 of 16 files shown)


vansin:
Follow-up on the three red checks:

1. `test-suite-registration`: true red. The new `tests/test1280-desktop-user-push` suite had a Dockerfile/run.sh but was not registered in CI. Fixed by adding it to both `qa.yml` path trigger lists and the `hub-orphan-gates` matrix. Verified only that gate:
   `sg docker -c 'docker run --rm -v "$PWD:/workspace" -w /workspace python:3.12-slim sh -lc "python3 .github/scripts/check-test-suite-registration.py --selftest && python3 .github/scripts/check-test-suite-registration.py"'`
   Result: `new=0`, exit 0.

2. `test629-mcp-schema-policy`: true red. `send_desktop_message` is a new production MCP tool, so the explicit schema/tool inventory needed to move from 46 to 47 and include the tool in source order before `broadcast`. Verified only that suite:
   `sg docker -c 'docker build --build-arg TEST629_SOURCE_COMMIT=$(git rev-parse HEAD) -t anet-test629-mcp-schema-policy -f tests/test629-mcp-schema-policy/Dockerfile . && docker run --rm anet-test629-mcp-schema-policy'`
   Result: mutation gates red, final probe green, `inventory_count=47`, `RESULT: PASS`.

3. `Windows Codex co-presence native smoke`: inspected artifact instead of rerunning. Failure signature is `EPERM: operation not permitted, rename ... .windows-copresence.json.<tmp> -> windows-copresence.json` during `node start windows-picker --no-inherit-codex-home` restart. The suite's own unit portion had `18 pass / 0 fail`. This does not intersect this PR's changed client/Hub SSE/message-count paths, so I am treating it as the known Windows runner flake unless a rerun reproduces with a different signature.

Message-count bug 判据:
- event path: client can receive `new_task` SSE carrying `inbox_count`; Chat now applies that immediately.
- data/render path: `/api/messages?alias=...` now actually filters to that alias and returns `pending_count`; Chat renders `pending_count > 0` as the visible new-message badge.
- regression test: `GET /api/messages — client inbox count` seeds unread/read/other-alias inbox rows and asserts `pending_count === 1` plus no cross-alias rows.

